PR #3, opened 2026-02-25
Add cache state data structures

Andrewtangtang:
## Summary

This PR add the data structures for the query condition cache extension:

- RowGroupBitvector: Per-row-group bitmap tracking which vectors
    contain qualifying rows. Uses `array<uint64_t, N>` where N is
    computed at compile time from `DEFAULT_ROW_GROUP_SIZE / STANDARD_VECTOR_SIZE`.

- ConditionCacheEntry: A single cache entry for a (table, predicate)
    pair. Uses table OID for lookup to survive renames within a session.
    Stores serialized filter key, a map of row group index to bitvector,
    and total qualifying row count.

- ConditionCacheStore: In-memory cache store backed by DuckDB's
    per-database `ObjectCache`. Supports lookup by filter key or table OID,
    insert, clear, and get-all operations, all protected by a mutex.

### DIFF
--- a/src/include/query_condition_cache_state.hpp
+++ b/src/include/query_condition_cache_state.hpp
@@ -8,69 +8,34 @@
 namespace duckdb {
 
 // Derived from DuckDB's compile-time configurable constants
-static constexpr idx_t VECTORS_PER_ROW_GROUP = DEFAULT_ROW_GROUP_SIZE / STANDARD_VECTOR_SIZE;
-static constexpr idx_t BITVECTOR_WORDS = (VECTORS_PER_ROW_GROUP + 63) / 64;
-static_assert(VECTORS_PER_ROW_GROUP > 0, "VECTORS_PER_ROW_GROUP must be positive");
+inline constexpr idx_t VECTORS_PER_ROW_GROUP = DEFAULT_ROW_GROUP_SIZE / STANDARD_VECTOR_SIZE;
+inline constexpr idx_t BITVECTOR_WORDS = (VECTORS_PER_ROW_GROUP + 63) / 64;
+static_assert(DEFAULT_ROW_GROUP_SIZE % STANDARD_VECTOR_SIZE == 0,
+              "DEFAULT_ROW_GROUP_SIZE must be divisible by STANDARD_VECTOR_SIZE");
 
 // Per row-group bitvector: bit[i] = 1 means vector i has at least one qualifying row,
 // for 0 <= i < VECTORS_PER_ROW_GROUP. Callers must not pass out-of-range indices;
+// D_ASSERT enforces this in debug builds only.
 // Backed by array<uint64_t, N> to support configurable row group / vector sizes.
 struct RowGroupBitvector {
 	array<uint64_t, BITVECTOR_WORDS> words = {};
 
-	bool VectorHasRows(idx_t vector_index) const {
-		D_ASSERT(vector_index < VECTORS_PER_ROW_GROUP);
-		return (words[vector_index / 64] >> (vector_index % 64)) & 1ULL;
-	}
-
-	void SetVector(idx_t vector_index) {
-		D_ASSERT(vector_index < VECTORS_PER_ROW_GROUP);
-		words[vector_index / 64] |= (1ULL << (vector_index % 64));
-	}
-
-	bool IsEmpty() const {
-		for (auto &w : words) {
-			if (w != 0) {
-				return false;
-			}
-		}
-		return true;
-	}
-
-	idx_t CountSetBits() const {
-		idx_t count = 0;
-		for (auto &w : words) {
-			count += static_cast<idx_t>(__builtin_popcountll(w));
-		}
-		return count;
-	}
+	bool VectorHasRows(idx_t vector_index) const;
+	void SetVector(idx_t vector_index);
+	bool IsEmpty() const;
+	idx_t CountSetBits() const;
 };
 
-// A single cache entry: the bitvectors for one (table, predicate) combination
+// A single cache entry: the bitvectors for one (table, predicate) combination.
+// table_oid is used for lookup within a session.
 struct ConditionCacheEntry {
-	string table_name;
+	idx_t table_oid;
 	string filter_key;
 	unordered_map<idx_t, RowGroupBitvector> bitvectors; // rg_idx -> bitvector
 	idx_t total_qualifying_rows = 0;
 };
 
-// Thread-local state for capturing filter key during query_condition_cache_build
-struct BuildCaptureState {
-	bool active = false;
-	string target_table;
-	string captured_filter_key;
-};
-
-// TODO: Replace thread_local variables with ClientContextState to avoid non-trivial thread_local types.
-
-// Global thread-local build capture state
-extern thread_local BuildCaptureState tl_build_capture;
-
-// Pre-optimize match: table_index -> cache entry to inject in the post-optimize pass.
-// Populated by PreOptimizeFunction, consumed and cleared by OptimizeFunction.
-extern thread_local unordered_map<idx_t, shared_ptr<ConditionCacheEntry>> tl_lookup_pending;
-
-// Stored in DuckDB's per-database ObjectCache, non-evictable
+// Stored in DuckDB's per-database ObjectCache
 class ConditionCacheStore : public ObjectCacheEntry {
 public:
 	static constexpr const char *CACHE_KEY = "query_condition_cache_store";
@@ -86,16 +51,16 @@ public:
 	// Lookup by filter key
 	shared_ptr<ConditionCacheEntry> Lookup(const string &filter_key);
 
-	// Lookup all entries for a table
-	vector<shared_ptr<ConditionCacheEntry>> LookupByTable(const string &table_name);
+	// Lookup all entries for a table by OID
+	vector<shared_ptr<ConditionCacheEntry>> LookupByTable(idx_t table_oid);
 
-	// Insert an entry
+	// Insert or update an entry
 	void Insert(shared_ptr<ConditionCacheEntry> entry);
 
 	// Clear all entries
 	void Clear();
 
-	// Get all entries (for query_condition_cache_info)
+	// Get all entries
 	vector<shared_ptr<ConditionCacheEntry>> GetAll();
 
 	// Get or create the store from a client context
--- a/src/include/query_condition_cache_state.hpp
+++ b/src/include/query_condition_cache_state.hpp
@@ -1,0 +1,112 @@
+#pragma once
+
+#include "duckdb/common/array.hpp"
+#include "duckdb/common/mutex.hpp"
+#include "duckdb/common/unordered_map.hpp"
+#include "duckdb/storage/object_cache.hpp"
+
+namespace duckdb {
+
+// Derived from DuckDB's compile-time configurable constants
+static constexpr idx_t VECTORS_PER_ROW_GROUP = DEFAULT_ROW_GROUP_SIZE / STANDARD_VECTOR_SIZE;
+static constexpr idx_t BITVECTOR_WORDS = (VECTORS_PER_ROW_GROUP + 63) / 64;
+
+// Per row-group bitvector: bit[i] = 1 means vector i has at least one qualifying row.
+// Backed by array<uint64_t, N> to support configurable row group / vector sizes.
+struct RowGroupBitvector {
+	array<uint64_t, BITVECTOR_WORDS> words = {};
+
+	bool VectorHasRows(idx_t vector_index) const {
+		D_ASSERT(vector_index < VECTORS_PER_ROW_GROUP);
+		return (words[vector_index / 64] >> (vector_index % 64)) & 1ULL;
+	}
+
+	void SetVector(idx_t vector_index) {
+		D_ASSERT(vector_index < VECTORS_PER_ROW_GROUP);
+		words[vector_index / 64] |= (1ULL << (vector_index % 64));
+	}
+
+	bool IsEmpty() const {
+		for (auto &w : words) {
+			if (w != 0) {
+				return false;
+			}
+		}
+		return true;
+	}
+
+	idx_t CountSetBits() const {
+		idx_t count = 0;
+		for (auto &w : words) {
+			uint64_t v = w;
+			while (v) {
+				count += v & 1ULL;
+				v >>= 1;
+			}
+		}
+		return count;
+	}
+};
+
+// A single cache entry: the bitvectors for one (table, predicate) combination
+struct ConditionCacheEntry {
+	string table_name;
+	string filter_key;
+	unordered_map<idx_t, RowGroupBitvector> bitvectors; // rg_idx -> bitvector
+	idx_t total_qualifying_rows = 0;
+};
+
+// Thread-local state for capturing filter key during query_condition_cache_build
+struct BuildCaptureState {
+	bool active = false;
+	string target_table;
+	string captured_filter_key;
+};
+
+// TODO: Replace thread_local variables with ClientContextState to avoid non-trivial thread_local types.
+
+// Global thread-local build capture state
+extern thread_local BuildCaptureState tl_build_capture;
+
+// Pre-optimize match: table_index -> cache entry to inject in the post-optimize pass.
+// Populated by PreOptimizeFunction, consumed and cleared by OptimizeFunction.
+extern thread_local unordered_map<idx_t, shared_ptr<ConditionCacheEntry>> tl_lookup_pending;
+
+// Stored in DuckDB's per-database ObjectCache, non-evictable
+class ConditionCacheStore : public ObjectCacheEntry {
+public:
+	static constexpr const char *CACHE_KEY = "query_condition_cache_store";
+
+	static string ObjectType() {
+		return "query_condition_cache_store";
+	}
+
+	string GetObjectType() override {
+		return ObjectType();
+	}
+
+	// Lookup by filter key
+	shared_ptr<ConditionCacheEntry> Lookup(const string &filter_key);
+
+	// Lookup all entries for a table
+	vector<shared_ptr<ConditionCacheEntry>> LookupByTable(const string &table_name);
+
+	// Insert an entry
+	void Insert(shared_ptr<ConditionCacheEntry> entry);
+
+	// Clear all entries
+	void Clear();
+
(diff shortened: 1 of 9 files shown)


src/include/query_condition_cache_state.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+#include "duckdb/common/array.hpp"
+#include "duckdb/common/mutex.hpp"
+#include "duckdb/common/unordered_map.hpp"
+#include "duckdb/storage/object_cache.hpp"
+
+namespace duckdb {
+
+// Derived from DuckDB's compile-time configurable constants
+static constexpr idx_t VECTORS_PER_ROW_GROUP = DEFAULT_ROW_GROUP_SIZE / STANDARD_VECTOR_SIZE;
```
dentiny: `inline constexpr`

src/include/query_condition_cache_state.hpp
```diff
@@ -0,0 +1,110 @@
...
+static constexpr idx_t BITVECTOR_WORDS = (VECTORS_PER_ROW_GROUP + 63) / 64;
+static_assert(VECTORS_PER_ROW_GROUP > 0, "VECTORS_PER_ROW_GROUP must be positive");
```
dentiny: not sure why it's needed, you use `idx_t` already

src/include/query_condition_cache_state.hpp
```diff
@@ -0,0 +1,110 @@
...
+				return false;
+			}
+		}
+		return true;
+	}
+
+	idx_t CountSetBits() const {
+		idx_t count = 0;
+		for (auto &w : words) {
+			count += static_cast<idx_t>(__builtin_popcountll(w));
```
dentiny: Please avoid using gcc-specific builtins

src/include/query_condition_cache_state.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+// Derived from DuckDB's compile-time configurable constants
+static constexpr idx_t VECTORS_PER_ROW_GROUP = DEFAULT_ROW_GROUP_SIZE / STANDARD_VECTOR_SIZE;
+static constexpr idx_t BITVECTOR_WORDS = (VECTORS_PER_ROW_GROUP + 63) / 64;
+static_assert(VECTORS_PER_ROW_GROUP > 0, "VECTORS_PER_ROW_GROUP must be positive");
+
+// Per row-group bitvector: bit[i] = 1 means vector i has at least one qualifying row,
+// for 0 <= i < VECTORS_PER_ROW_GROUP. Callers must not pass out-of-range indices;
+// Backed by array<uint64_t, N> to support configurable row group / vector sizes.
+struct RowGroupBitvector {
```
dentiny: I would suggest place declaration in header files, and move implementation to impl file, so we could potentially do forward declaration and less dependency in header file to improve compile time

src/include/query_condition_cache_state.hpp
```diff
@@ -0,0 +1,110 @@
...
+	string target_table;
+	string captured_filter_key;
+};
+
+// TODO: Replace thread_local variables with ClientContextState to avoid non-trivial thread_local types.
+
+// Global thread-local build capture state
+extern thread_local BuildCaptureState tl_build_capture;
+
+// Pre-optimize match: table_index -> cache entry to inject in the post-optimize pass.
```
dentiny: could you please explain to me a bit what's "table index"? I'm not too familiar with it, thanks!

src/include/query_condition_cache_state.hpp
```diff
@@ -0,0 +1,110 @@
...
+		for (auto &w : words) {
+			count += static_cast<idx_t>(__builtin_popcountll(w));
+		}
+		return count;
+	}
+};
+
+// A single cache entry: the bitvectors for one (table, predicate) combination
+struct ConditionCacheEntry {
+	string table_name;
```
dentiny: should we use fully qualified table name?

dentiny: also, how do you consider table and database rename? should we use OID since users could rename entities?

src/include/query_condition_cache_state.hpp
dentiny: static assert there's no mod

src/query_condition_cache_state.cpp
```diff
@@ -0,0 +1,56 @@
+#include "query_condition_cache_state.hpp"
+
+namespace duckdb {
+
+// Thread-local build capture state
+thread_local BuildCaptureState tl_build_capture;
```
dentiny: these two thread-locals are not used, generally I don't want to checkin unused code;
could you please either remove or leave a TODO so at least we could trace back and check before merge?

src/query_condition_cache_state.cpp
```diff
@@ -0,0 +1,56 @@
...
+}
+
+void ConditionCacheStore::Clear() {
+	lock_guard<mutex> guard(cache_lock);
+	entries.clear();
+}
+
+vector<shared_ptr<ConditionCacheEntry>> ConditionCacheStore::GetAll() {
+	lock_guard<mutex> guard(cache_lock);
+	vector<shared_ptr<ConditionCacheEntry>> result;
```
dentiny, line 56: why put memory allocation inside of critical section?

src/query_condition_cache_state.cpp
```diff
@@ -0,0 +1,56 @@
...
+		if (pair.second->table_name == table_name) {
+			result.push_back(pair.second);
+		}
+	}
+	return result;
+}
+
+void ConditionCacheStore::Insert(shared_ptr<ConditionCacheEntry> entry) {
+	lock_guard<mutex> guard(cache_lock);
+	entries[entry->filter_key] = std::move(entry);
```
dentiny: could you use APIs like `emplace` to check return value?

dentiny (commented):
Quick first pass... It reads like two PRs -- I don't see any unit test or usage for filter key related, thread-local variables also not used

dentiny (commented):
We also need to implement LRU and max memory cap, right?

Andrewtangtang:
I think we should. I'll create an issue to track this. Or do you have a preferred strategy for this?

src/include/query_condition_cache_state.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+#include "duckdb/common/array.hpp"
+#include "duckdb/common/mutex.hpp"
+#include "duckdb/common/unordered_map.hpp"
+#include "duckdb/storage/object_cache.hpp"
+
+namespace duckdb {
+
+// Derived from DuckDB's compile-time configurable constants
+inline constexpr idx_t VECTORS_PER_ROW_GROUP = DEFAULT_ROW_GROUP_SIZE / STANDARD_VECTOR_SIZE;
```
dentiny, line 12: I suspect you need to add some [compilation options](https://github.com/dentiny/duck-read-cache-fs/blob/d3618a6fd8654db431f0f357cd08c89035411ad6/CMakeLists.txt#L10C46-L10C89) otherwise default C++11 will issue warnings.

BTW warning on unused function / variables is useful.

src/include/query_condition_cache_state.hpp
```diff
@@ -0,0 +1,75 @@
...
+	void SetVector(idx_t vector_index);
+	bool IsEmpty() const;
+	idx_t CountSetBits() const;
+};
+
+// A single cache entry: the bitvectors for one (table, predicate) combination.
+// table_oid is used for lookup within a session.
+struct ConditionCacheEntry {
+	idx_t table_oid;
+	string filter_key;
```
dentiny, line 36: Curious why do we filter key here? I thought it's used to lookup cache entry.
Asking because I expect it to be serialized query thus won't be small.

src/query_condition_cache_state.cpp
```diff
@@ -0,0 +1,90 @@
...
+void ConditionCacheStore::Clear() {
+	lock_guard<mutex> guard(cache_lock);
+	entries.clear();
+}
+
+vector<shared_ptr<ConditionCacheEntry>> ConditionCacheStore::GetAll() {
+	vector<shared_ptr<ConditionCacheEntry>> result;
+	{
+		lock_guard<mutex> guard(cache_lock);
+		result.reserve(entries.size());
```
dentiny, line 59: Ahhh sorry I mis-read last time

src/include/query_condition_cache_state.hpp
```diff
@@ -0,0 +1,75 @@
...
+	// Get all entries
+	vector<shared_ptr<ConditionCacheEntry>> GetAll();
+
+	// Get or create the store from a client context
+	static shared_ptr<ConditionCacheStore> GetOrCreate(ClientContext &context);
+
+private:
+	mutex cache_lock;
+	// filter_key -> entry
+	unordered_map<string, shared_ptr<ConditionCacheEntry>> entries;
```
dentiny: Let's leave a TODO for sharding.

src/include/query_condition_cache_state.hpp
```diff
@@ -0,0 +1,75 @@
...
+#include "duckdb/common/array.hpp"
+#include "duckdb/common/mutex.hpp"
+#include "duckdb/common/unordered_map.hpp"
+#include "duckdb/storage/object_cache.hpp"
+
+namespace duckdb {
+
+// Derived from DuckDB's compile-time configurable constants
+inline constexpr idx_t VECTORS_PER_ROW_GROUP = DEFAULT_ROW_GROUP_SIZE / STANDARD_VECTOR_SIZE;
+inline constexpr idx_t BITVECTOR_WORDS = (VECTORS_PER_ROW_GROUP + 63) / 64;
```
dentiny: curious how should I understand "WORDS" here?

src/include/query_condition_cache_state.hpp
```diff
@@ -0,0 +1,75 @@
...
+static_assert(DEFAULT_ROW_GROUP_SIZE % STANDARD_VECTOR_SIZE == 0,
+              "DEFAULT_ROW_GROUP_SIZE must be divisible by STANDARD_VECTOR_SIZE");
+
+// Per row-group bitvector: bit[i] = 1 means vector i has at least one qualifying row,
+// for 0 <= i < VECTORS_PER_ROW_GROUP. Callers must not pass out-of-range indices;
```
dentiny: we could do `.at()` and remove the debug assertion

src/query_condition_cache_state.cpp
```diff
@@ -0,0 +1,90 @@
...
+		return it->second;
+	}
+	return nullptr;
+}
+
+vector<shared_ptr<ConditionCacheEntry>> ConditionCacheStore::LookupByTable(idx_t table_oid) {
+	lock_guard<mutex> guard(cache_lock);
+	vector<shared_ptr<ConditionCacheEntry>> result;
+	for (auto &pair : entries) {
+		if (pair.second->table_oid == table_oid) {
```
dentiny: table number could be big, how do you plan to address?

src/include/query_condition_cache_state.hpp
Andrewtangtang: I plan to use the filter_key as a sorted concatenation of expression strings (e.g. 42|cat=42;val>5), not the full query, so I think it wouldn't be very large. That said, it's currently only used as the map key during Insert, so I've removed it.

src/include/query_condition_cache_state.hpp
Andrewtangtang: I have replaced it with "BITVECTOR_ARRAY_SIZE".

src/query_condition_cache_state.cpp
Andrewtangtang: I added another `unordered_map` "entries_by_table" (table_oid -> filter_keys) for efficient lookup

src/query_condition_cache_state.cpp
```diff
@@ -0,0 +1,97 @@
...
+	}
+	return true;
+}
+
+idx_t RowGroupBitvector::CountSetBits() const {
+	idx_t count = 0;
+	for (const auto &w : matching_vectors) {
+		uint64_t v = w;
+		while (v) {
+			v &= v - 1;
```
dentiny:
```suggestion
			v &= (v - 1);
```

src/include/query_condition_cache_state.hpp
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	string GetObjectType() override {
+		return ObjectType();
+	}
+
+	// Lookup by filter key
+	shared_ptr<ConditionCacheEntry> Lookup(const string &filter_key);
+
+	// Lookup all entries for a table by OID
```
dentiny: need documentation on what if requested key doesn't exist, same below

src/include/query_condition_cache_state.hpp
```diff
@@ -0,0 +1,75 @@
...
+	vector<shared_ptr<ConditionCacheEntry>> LookupByTable(idx_t table_oid);
+
+	// Insert or update an entry
```
dentiny:
```suggestion
// Upsert an entry
void Upsert(const string &filter_key, shared_ptr<ConditionCacheEntry> entry);
```

src/include/query_condition_cache_state.hpp
```diff
@@ -0,0 +1,75 @@
...
+	vector<shared_ptr<ConditionCacheEntry>> GetAll();
+
+	// Get or create the store from a client context
+	static shared_ptr<ConditionCacheStore> GetOrCreate(ClientContext &context);
+
+private:
+	mutex cache_lock;
+	// TODO: Consider sharding for scalability
+	// filter_key -> entry
+	unordered_map<string, shared_ptr<ConditionCacheEntry>> entries;
```
dentiny: Emmm why do we need two maps here? how do you plan to do LRU later?
I think whether on write or read, the query key could both be <table, query>
so the key could be
```cpp
struct CacheKey {
 idx_t oid;
 string key;
};
```
?

src/include/query_condition_cache_state.hpp
```diff
@@ -0,0 +1,75 @@
...
+// Per row-group bitvector: bit[i] = 1 means vector i has at least one qualifying row,
+// for 0 <= i < VECTORS_PER_ROW_GROUP.
+// Backed by array<uint64_t, N> to support configurable row group / vector sizes.
+struct RowGroupBitvector {
+	array<uint64_t, BITVECTOR_ARRAY_SIZE> matching_vectors = {};
+
+	bool VectorHasRows(idx_t vector_index) const;
+	void SetVector(idx_t vector_index);
+	bool IsEmpty() const;
+	idx_t CountSetBits() const;
```
dentiny: Do you think it's better if we rename it to high-level abstractions?
So bitmap is a detailed implementation, which we could switch to other data structure like roaring bitmap, map, etc.

src/include/query_condition_cache_state.hpp
```diff
@@ -0,0 +1,75 @@
...
+// Derived from DuckDB's compile-time configurable constants
+inline constexpr idx_t VECTORS_PER_ROW_GROUP = DEFAULT_ROW_GROUP_SIZE / STANDARD_VECTOR_SIZE;
+inline constexpr idx_t BITVECTOR_ARRAY_SIZE = (VECTORS_PER_ROW_GROUP + 63) / 64;
+static_assert(DEFAULT_ROW_GROUP_SIZE % STANDARD_VECTOR_SIZE == 0,
+              "DEFAULT_ROW_GROUP_SIZE must be divisible by STANDARD_VECTOR_SIZE");
+
+// Per row-group bitvector: bit[i] = 1 means vector i has at least one qualifying row,
+// for 0 <= i < VECTORS_PER_ROW_GROUP.
+// Backed by array<uint64_t, N> to support configurable row group / vector sizes.
+struct RowGroupBitvector {
```
dentiny: Is it possible that `RowGroupBitvector` is updated by two threads?
i.e., two thread executes the same (sub)query on the same row group?

src/include/query_condition_cache_state.hpp
dentiny: Curious when we plan to use `CountSetBits`?

dentiny (commented):
comment left

src/include/query_condition_cache_state.hpp
Andrewtangtang: How about `RowGroupFilter` or `RowGroupMask`

Andrewtangtang: I originally added this function for observability, but it’s not really useful right now, so I’m just removing it for now.

src/include/query_condition_cache_state.hpp
Andrewtangtang: For now I plan to have users manually build the cache (like `condition_cache_build()`) first to utilize the cache, so `SetVector()` is only called from a single thread to create. We can revisit if we later support automatic cache population.

dentiny: Then it should be constructor

src/include/query_condition_cache_state.hpp
```diff
@@ -16,20 +17,44 @@ static_assert(DEFAULT_ROW_GROUP_SIZE % STANDARD_VECTOR_SIZE == 0,
...
 	bool IsEmpty() const;
-	idx_t CountSetBits() const;
+};
+
+// Composite key for cache lookup: (table_oid, filter_key)
+struct CacheKey {
+	idx_t table_oid;
+	string filter_key;
+
+	bool operator==(const CacheKey &other) const {
```
dentiny, line 38: If you define equal struct, we don't need equality comparison then?

src/query_condition_cache_state.cpp
```diff
@@ -21,60 +23,28 @@ bool RowGroupBitvector::IsEmpty() const {
...
-			if (entry_it != entries.end()) {
-				result.push_back(entry_it->second);
-			}
-		}
-	}
-	return result;
-}
-
-void ConditionCacheStore::Insert(const string &filter_key, shared_ptr<ConditionCacheEntry> entry) {
+void ConditionCacheStore::Upsert(const CacheKey &key, shared_ptr<ConditionCacheEntry> entry) {
```
dentiny, line 39: let's check entry not null, otherwise throw invalid input exception

dentiny (commented):
LG! just two minor comments

dentiny (approved):
Thanks!

Andrewtangtang:
Thanks for reviewing.

Issue #5, opened 2026-03-03
Canonical predicate key for cache lookup


Andrewtangtang:
**Is your feature request related to a problem? Please describe.**

The cache key currently uses the raw predicate string. `col1 > 5 AND col2 > 10` and `col2 > 10 AND col1 > 5` produce different keys but are semantically identical, causing cache misses.

**Describe the solution you'd like**

Generate a canonical key from the optimizer's internal representation instead of the raw SQL string. Two approaches:
- For simple filters pushed down to `TableFilterSet`: serialize by column index order 
- For complex filters in `LogicalFilter`: serialize the `Expression` tree in a canonical form

Both build and lookup should use the same serialization logic so keys always match.

**Describe alternatives you've considered**

N/A

**Additional context**

N/A

Issue #4, opened 2026-03-03
Cache invalidation on table modification


Andrewtangtang:
**Is your feature request related to a problem? Please describe.**

Any table modification (INSERT/DELETE/UPDATE/VACUUM) can make cached rowids stale. Currently there is no mechanism to detect or invalidate stale cache entries.

**Describe the solution you'd like**

- Manual invalidation via a `condition_cache_clear` function or get version information when query(need to investigate).

**Describe alternatives you've considered**

N/A

**Additional context**

N/A

PR #6, opened 2026-03-03
Add condition_cache_build table function

Andrewtangtang:
## Summary

This PR adds a table function `condition_cache_build(table_name, predicate_sql)` for users to create the condition cache manually. Internally, the table is resolved by OID and supports qualified table names (`schema.table`, `catalog.schema.table`). Cache was builded by using storage-layer APIs directly. 

## Note

This PR has two open TODOs to be addressed in the future:
- Cache invalidation mechanism — see #4 
- Canonical predicate key extraction — see #5

### DIFF
--- a/src/include/query_condition_cache_functions.hpp
+++ b/src/include/query_condition_cache_functions.hpp
@@ -1,0 +1,20 @@
+#pragma once
+
+#include "query_condition_cache_state.hpp"
+
+#include "duckdb/function/table_function.hpp"
+
+namespace duckdb {
+
+class DuckTableEntry;
+class Expression;
+
+// Scan all rows in the table, evaluate bound_expr, and build a ConditionCacheEntry
+// recording which vectors contain qualifying rows.
+// Modifies bound_expr in-place (remaps column indices to scan positions).
+shared_ptr<ConditionCacheEntry> BuildCacheEntry(ClientContext &context, DuckTableEntry &table_entry,
+                                                Expression &bound_expr, idx_t &total_qualifying_rows);
+
+TableFunction ConditionCacheBuildFunction();
+
+} // namespace duckdb
--- a/src/include/query_condition_cache_functions.hpp
+++ b/src/include/query_condition_cache_functions.hpp
@@ -14,7 +14,7 @@ class Expression;
 // recording which vectors contain qualifying rows.
 // Modifies bound_expr in-place (remaps column indices to scan positions).
 shared_ptr<ConditionCacheEntry> BuildCacheEntry(ClientContext &context, DuckTableEntry &table_entry,
-                                                Expression &bound_expr, idx_t &total_qualifying_rows);
+                                                Expression &bound_expr);
 
 struct CacheEntryStats {
 	idx_t qualifying_vectors;
--- a/src/include/query_condition_cache_functions.hpp
+++ b/src/include/query_condition_cache_functions.hpp
@@ -9,11 +9,22 @@ namespace duckdb {
 class DuckTableEntry;
 class Expression;
 
+// TODO: Exposed for future reuse and C++ unit testing.
 // Scan all rows in the table, evaluate bound_expr, and build a ConditionCacheEntry
 // recording which vectors contain qualifying rows.
 // Modifies bound_expr in-place (remaps column indices to scan positions).
 shared_ptr<ConditionCacheEntry> BuildCacheEntry(ClientContext &context, DuckTableEntry &table_entry,
                                                 Expression &bound_expr, idx_t &total_qualifying_rows);
+
+struct CacheEntryStats {
+	idx_t qualifying_vectors;
+	idx_t total_vectors;
+	idx_t qualifying_row_groups;
+	idx_t total_row_groups;
+};
+
+// TODO: Exposed for future reuse. 
+CacheEntryStats ComputeCacheEntryStats(const ConditionCacheEntry &entry, idx_t total_rows);
 
 TableFunction ConditionCacheBuildFunction();
 
--- a/src/include/query_condition_cache_functions.hpp
+++ b/src/include/query_condition_cache_functions.hpp
@@ -23,7 +23,7 @@ struct CacheEntryStats {
 	idx_t total_row_groups;
 };
 
-// TODO: Exposed for future reuse. 
+// TODO: Exposed for future reuse.
 CacheEntryStats ComputeCacheEntryStats(const ConditionCacheEntry &entry, idx_t total_rows);
 
 TableFunction ConditionCacheBuildFunction();
--- a/src/query_condition_cache_extension.cpp
+++ b/src/query_condition_cache_extension.cpp
@@ -1,14 +1,11 @@
 #define DUCKDB_EXTENSION_MAIN
 
 #include "query_condition_cache_extension.hpp"
+#include "query_condition_cache_functions.hpp"
 
-#include "duckdb/function/table_function.hpp"
 #include "duckdb/main/extension/extension_loader.hpp"
 
 namespace duckdb {
-
-// Defined in query_condition_cache_functions.cpp
-TableFunction ConditionCacheBuildFunction();
 
 namespace {
 void LoadInternal(ExtensionLoader &loader) {
--- a/src/query_condition_cache_extension.cpp
+++ b/src/query_condition_cache_extension.cpp
@@ -2,13 +2,17 @@
 
 #include "query_condition_cache_extension.hpp"
 
+#include "duckdb/function/table_function.hpp"
 #include "duckdb/main/extension/extension_loader.hpp"
 
 namespace duckdb {
 
+// Defined in query_condition_cache_functions.cpp
+TableFunction ConditionCacheBuildFunction();
+
 namespace {
 void LoadInternal(ExtensionLoader &loader) {
-	// Functions and optimizer will be registered in subsequent PRs.
+	loader.RegisterFunction(ConditionCacheBuildFunction());
 }
 } // namespace
 
--- a/src/query_condition_cache_extension.cpp
+++ b/src/query_condition_cache_extension.cpp
@@ -1,9 +1,9 @@
 #define DUCKDB_EXTENSION_MAIN
 
 #include "query_condition_cache_extension.hpp"
-#include "query_condition_cache_functions.hpp"
 
 #include "duckdb/main/extension/extension_loader.hpp"
+#include "query_condition_cache_functions.hpp"
 
 namespace duckdb {
 
--- a/src/query_condition_cache_functions.cpp
+++ b/src/query_condition_cache_functions.cpp
@@ -33,7 +33,7 @@ struct ConditionCacheBuildBindData : public TableFunctionData {
 	string table;
 	string predicate_sql;
 	idx_t table_oid;
-	idx_t total_rows;
+	idx_t total_rows; // total number of rows in the table
 };
 
 struct ConditionCacheBuildState : public GlobalTableFunctionState {
@@ -48,9 +48,9 @@ unique_ptr<FunctionData> ConditionCacheBuildBind(ClientContext &context, TableFu
 	// Parse qualified table name (supports "table", "schema.table", "catalog.schema.table")
 	auto qname = QualifiedName::Parse(input.inputs[0].GetValue<string>());
 	Binder::BindSchemaOrCatalog(context, qname.catalog, qname.schema);
-	result->catalog = qname.catalog;
-	result->schema = qname.schema;
-	result->table = qname.name;
+	result->catalog = std::move(qname.catalog);
+	result->schema = std::move(qname.schema);
+	result->table = std::move(qname.name);
 
 	auto &table_entry = Catalog::GetEntry<DuckTableEntry>(context, result->catalog, result->schema, result->table);
 	result->table_oid = table_entry.oid;
@@ -66,20 +66,32 @@ unique_ptr<GlobalTableFunctionState> ConditionCacheBuildInit(ClientContext &cont
 	return make_uniq<ConditionCacheBuildState>();
 }
 
+// Collect all column OIDs referenced by bound expressions (BoundReferenceExpression).
+void CollectReferencedColumns(Expression &expr, unordered_set<column_t> &column_oids) {
+	if (expr.GetExpressionClass() == ExpressionClass::BOUND_REF) {
+		auto &ref = expr.Cast<BoundReferenceExpression>();
+		column_oids.insert(ref.index);
+	}
+	ExpressionIterator::EnumerateChildren(expr,
+	                                      [&](Expression &child) { CollectReferencedColumns(child, column_oids); });
+}
+
 // After binding, column indices refer to StorageOid which may not match the scan order.
 // Remap them to the actual DataChunk column positions.
-void RemapColumnIndices(Expression &expr, const unordered_map<idx_t, idx_t> &mapping) {
+void RemapColumnIndices(Expression &expr, const unordered_map<column_t, idx_t> &mapping) {
 	if (expr.GetExpressionClass() == ExpressionClass::BOUND_REF) {
 		auto &ref = expr.Cast<BoundReferenceExpression>();
 		auto it = mapping.find(ref.index);
-		if (it != mapping.end()) {
-			ref.index = it->second;
-		}
+		D_ASSERT(it != mapping.end());
+		ref.index = it->second;
 	}
-	ExpressionIterator::EnumerateChildren(expr, [&](Expression &child) {
-		RemapColumnIndices(child, mapping);
-	});
+	ExpressionIterator::EnumerateChildren(expr, [&](Expression &child) { RemapColumnIndices(child, mapping); });
 }
+
+struct ScanColumn {
+	StorageIndex index;
+	LogicalType type;
+};
 
 } // namespace
 
@@ -90,23 +102,34 @@ shared_ptr<ConditionCacheEntry> BuildCacheEntry(ClientContext &context, DuckTabl
 	auto &storage = table_entry.GetStorage();
 	auto &columns = table_entry.GetColumns();
 
-	// Scan all physical columns (needed by the predicate expression) plus rowid
-	vector<pair<StorageIndex, LogicalType>> scan_columns;
-	unordered_map<idx_t, idx_t> storage_to_scan_idx;
+	// Only scan columns referenced by the predicate, plus rowid
+	unordered_set<column_t> referenced_oids;
+	CollectReferencedColumns(bound_expr, referenced_oids);
+
+	vector<ScanColumn> scan_columns;
+	// Maps storage column OID to scan position in DataChunk
+	unordered_map<column_t, idx_t> storage_to_scan_idx;
+
+	scan_columns.reserve(referenced_oids.size() + 1);
 
 	idx_t scan_pos = 0;
-	for (auto &col : columns.Physical()) {
-		scan_columns.emplace_back(StorageIndex(col.Oid()), col.Type());
+	for (const auto &col : columns.Physical()) {
+		if (referenced_oids.count(col.Oid()) == 0) {
+			continue;
+		}
+		scan_columns.push_back({StorageIndex(col.Oid()), col.Type()});
 		storage_to_scan_idx[col.Oid()] = scan_pos++;
 	}
 	const idx_t rowid_col_idx = scan_pos;
-	scan_columns.emplace_back(StorageIndex(COLUMN_IDENTIFIER_ROW_ID), LogicalType::ROW_TYPE);
+	scan_columns.push_back({StorageIndex(COLUMN_IDENTIFIER_ROW_ID), LogicalType {LogicalType::ROW_TYPE}});
 
 	vector<StorageIndex> column_ids;
 	vector<LogicalType> scan_types;
-	for (const auto &[id, type] : scan_columns) {
-		column_ids.push_back(id);
-		scan_types.push_back(type);
+	column_ids.reserve(scan_columns.size());
+	scan_types.reserve(scan_columns.size());
+	for (const auto &sc : scan_columns) {
+		column_ids.push_back(sc.index);
+		scan_types.push_back(sc.type);
 	}
 
 	RemapColumnIndices(bound_expr, storage_to_scan_idx);
@@ -119,7 +142,7 @@ shared_ptr<ConditionCacheEntry> BuildCacheEntry(ClientContext &context, DuckTabl
 	// Scan and evaluate predicate
 	ExpressionExecutor executor(context, bound_expr);
 
-	unordered_map<idx_t, unordered_set<idx_t>> row_group_vec_indices;
+	auto entry = make_shared_ptr<ConditionCacheEntry>();
 
 	while (true) {
 		DataChunk chunk;
@@ -137,24 +160,18 @@ shared_ptr<ConditionCacheEntry> BuildCacheEntry(ClientContext &context, DuckTabl
 		rowid_vec.ToUnifiedFormat(chunk.size(), rowid_data);
 		auto rowids = UnifiedVectorFormat::GetData<row_t>(rowid_data);
 
-		for (idx_t i = 0; i < match_count; ++i) {
-			auto rid_idx = rowid_data.sel->get_index(sel.get_index(i));
-			if (!rowid_data.validity.RowIsValid(rid_idx)) {
+		for (idx_t idx = 0; idx < match_count; ++idx) {
+			auto rowid_entry = rowid_data.sel->get_index(sel.get_index(idx));
+			if (!rowid_data.validity.RowIsValid(rowid_entry)) {
 				continue;
 			}
-			auto row_id = NumericCast<idx_t>(rowids[rid_idx]);
+			auto row_id = NumericCast<idx_t>(rowids[rowid_entry]);
 			idx_t row_group_idx = row_id / DEFAULT_ROW_GROUP_SIZE;
 			idx_t vector_idx = (row_id % DEFAULT_ROW_GROUP_SIZE) / STANDARD_VECTOR_SIZE;
-			row_group_vec_indices[row_group_idx].insert(vector_idx);
+			entry->bitvectors[row_group_idx].SetVector(vector_idx);
 		}
 	}
 
-	// Build cache entry from collected indices
-	auto entry = make_shared_ptr<ConditionCacheEntry>();
-	for (const auto &pair : row_group_vec_indices) {
-		vector<idx_t> indices(pair.second.begin(), pair.second.end());
-		entry->bitvectors.emplace(pair.first, RowGroupFilter(indices));
-	}
 	return entry;
 }
 
@@ -180,7 +197,12 @@ CacheEntryStats ComputeCacheEntryStats(const ConditionCacheEntry &entry, idx_t t
 	idx_t qualifying_row_groups = entry.bitvectors.size();
 	idx_t total_row_groups = (total_rows + DEFAULT_ROW_GROUP_SIZE - 1) / DEFAULT_ROW_GROUP_SIZE;
 
-	return {qualifying_vectors, total_vectors, qualifying_row_groups, total_row_groups};
+	CacheEntryStats stats;
+	stats.qualifying_vectors = qualifying_vectors;
+	stats.total_vectors = total_vectors;
+	stats.qualifying_row_groups = qualifying_row_groups;
+	stats.total_row_groups = total_row_groups;
+	return stats;
 }
 
 void ConditionCacheBuildExecute(ClientContext &context, TableFunctionInput &data_p, DataChunk &output) {
@@ -211,12 +233,9 @@ void ConditionCacheBuildExecute(ClientContext &context, TableFunctionInput &data
 	CheckBinder check_binder(*binder, context, bind_data.table, table_entry.GetColumns(), bound_columns);
 	auto bound_expr = check_binder.Bind(parsed_exprs[0]);
 
-	// CheckBinder sets target_type = INTEGER, so all expressions get cast to INTEGER.
-	// Re-cast to BOOLEAN for SelectExpression.
-	if (bound_expr->return_type.id() != LogicalTypeId::BOOLEAN) {
-		bound_expr = BoundCastExpression::AddCastToType(context, std::move(bound_expr),
-		                                               LogicalType {LogicalTypeId::BOOLEAN});
-	}
+	// CheckBinder always sets target_type = INTEGER, so re-cast to BOOLEAN for SelectExpression.
+	bound_expr =
+	    BoundCastExpression::AddCastToType(context, std::move(bound_expr), LogicalType {LogicalTypeId::BOOLEAN});
 
 	auto entry = BuildCacheEntry(context, table_entry, *bound_expr);
 	auto stats = ComputeCacheEntryStats(*entry, bind_data.total_rows);
@@ -229,17 +248,16 @@ void ConditionCacheBuildExecute(ClientContext &context, TableFunctionInput &data
 	store->Upsert(key, std::move(entry));
 
 	output.SetCardinality(1);
-	output.data[0].SetValue(
-	    0, StringUtil::Format("Cache Built: %llu/%llu vectors, %llu/%llu row groups",
-	                          stats.qualifying_vectors, stats.total_vectors,
-	                          stats.qualifying_row_groups, stats.total_row_groups));
+	output.data[0].SetValue(0, StringUtil::Format("Cache Built: %llu/%llu vectors, %llu/%llu row groups",
+	                                              stats.qualifying_vectors, stats.total_vectors,
+	                                              stats.qualifying_row_groups, stats.total_row_groups));
 }
 
 TableFunction ConditionCacheBuildFunction() {
-	TableFunction func("condition_cache_build",
-	                   {LogicalType {LogicalTypeId::VARCHAR} /*table*/,
-	                    LogicalType {LogicalTypeId::VARCHAR} /*predicate_sql*/},
-	                   ConditionCacheBuildExecute, ConditionCacheBuildBind, ConditionCacheBuildInit);
+	TableFunction func(
+	    "condition_cache_build",
+	    {LogicalType {LogicalTypeId::VARCHAR} /*table*/, LogicalType {LogicalTypeId::VARCHAR} /*predicate_sql*/},
+	    ConditionCacheBuildExecute, ConditionCacheBuildBind, ConditionCacheBuildInit);
 	return func;
 }
 
--- a/src/query_condition_cache_functions.cpp
+++ b/src/query_condition_cache_functions.cpp
@@ -197,12 +197,12 @@ CacheEntryStats ComputeCacheEntryStats(const ConditionCacheEntry &entry, idx_t t
 	idx_t qualifying_row_groups = entry.bitvectors.size();
 	idx_t total_row_groups = (total_rows + DEFAULT_ROW_GROUP_SIZE - 1) / DEFAULT_ROW_GROUP_SIZE;
 
-	CacheEntryStats stats;
-	stats.qualifying_vectors = qualifying_vectors;
-	stats.total_vectors = total_vectors;
-	stats.qualifying_row_groups = qualifying_row_groups;
-	stats.total_row_groups = total_row_groups;
-	return stats;
+	return CacheEntryStats {
+	    .qualifying_vectors = qualifying_vectors,
+	    .total_vectors = total_vectors,
+	    .qualifying_row_groups = qualifying_row_groups,
+	    .total_row_groups = total_row_groups,
+	};
 }
 
 void ConditionCacheBuildExecute(ClientContext &context, TableFunctionInput &data_p, DataChunk &output) {
--- a/src/query_condition_cache_functions.cpp
+++ b/src/query_condition_cache_functions.cpp
@@ -1,5 +1,4 @@
 #include "query_condition_cache_functions.hpp"
-#include "query_condition_cache_state.hpp"
 
 #include "duckdb/catalog/catalog.hpp"
 #include "duckdb/catalog/catalog_entry/duck_table_entry.hpp"
@@ -21,6 +20,7 @@
 #include "duckdb/storage/storage_index.hpp"
 #include "duckdb/storage/table/scan_state.hpp"
 #include "duckdb/transaction/duck_transaction.hpp"
+#include "query_condition_cache_state.hpp"
 
 namespace duckdb {
 namespace {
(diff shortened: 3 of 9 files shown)


dentiny (commented):
Why do we need to manually? I thought everything should be automatically handled inside of duckdb and the extension?
Is it for testing?

Andrewtangtang:
Everything can be done automatically, but that needs further implementation in a future PR. This PR is for testing and also provides a way for users  who want to build their cache explicitly.

src/query_condition_cache_extension.cpp
```diff
@@ -2,13 +2,17 @@
 
 #include "query_condition_cache_extension.hpp"
 
+#include "duckdb/function/table_function.hpp"
 #include "duckdb/main/extension/extension_loader.hpp"
 
 namespace duckdb {
 
+// Defined in query_condition_cache_functions.cpp
+TableFunction ConditionCacheBuildFunction();
```
dentiny: We could also declare it in a separate header file, it won't increase too much of our compile/link overhead.
I rarely see people relying on linked symbols.

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,152 @@
+#include "query_condition_cache_state.hpp"
+
+#include "duckdb/catalog/catalog.hpp"
+#include "duckdb/catalog/catalog_entry/duck_table_entry.hpp"
+#include "duckdb/common/exception.hpp"
+#include "duckdb/common/string_util.hpp"
+#include "duckdb/common/unordered_map.hpp"
```
dentiny, line 8: alphabetical order?

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,152 @@
...
+};
+
+struct ConditionCacheBuildState : public GlobalTableFunctionState {
+	bool done = false;
+};
+
+unique_ptr<FunctionData> ConditionCacheBuildBind(ClientContext &context, TableFunctionBindInput &input,
+                                                 vector<LogicalType> &return_types, vector<string> &names) {
+	auto result = make_uniq<ConditionCacheBuildBindData>();
+	result->table_name = input.inputs[0].GetValue<string>();
```
dentiny: Should we check input arguments number?

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,152 @@
...
+	    0, StringUtil::Format("Cache Built: %llu qualifying rows, %llu/%llu vectors, %llu/%llu row groups",
+	                          total_qualifying_rows, qualifying_vectors, total_vectors, qualifying_row_groups,
+	                          total_row_groups));
+}
+
+} // namespace
+
+TableFunction ConditionCacheBuildFunction() {
+	TableFunction func("condition_cache_build",
+	                   {LogicalType {LogicalTypeId::VARCHAR}, LogicalType {LogicalTypeId::VARCHAR}},
```
dentiny: nit: I'd prefer if we could leave a comment besides the argument.
For example, `/*table=*/`

dentiny: Also, we should be able to handle `<database>.<schema>.<object>`, right?
Should we add a sql test to verify?
Also I think the object is not only tables, but could be indexes, right?

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,152 @@
...
+	idx_t qualifying_vectors = 0;
+	for (const auto &bitvector_pair : entry->bitvectors) {
+		for (idx_t v = 0; v < vectors_per_row_group; ++v) {
+			if (bitvector_pair.second.VectorHasRows(v)) {
+				++qualifying_vectors;
+			}
+		}
+	}
+
+	idx_t full_row_groups = bind_data.total_rows / DEFAULT_ROW_GROUP_SIZE;
```
dentiny: `divmod`? https://en.cppreference.com/w/cpp/numeric/math/div.html

test/sql/condition_cache_build.test
```diff
@@ -0,0 +1,39 @@
...
+
+require query_condition_cache
+
+# Create table with multiple row groups (>122880 rows)
+statement ok
+CREATE TABLE t AS SELECT i AS id, i % 100 AS val FROM range(500000) t(i);
+
+# Build cache for simple equality predicate
+query I
+SELECT status FROM condition_cache_build('t', 'val = 42');
```
dentiny, line 13: I'm curious how do we plan to handle cases like
A: id = 10 && val < 10
B: val < 10 && id = 10
Do these two predicates have the same serialized predicate?

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,152 @@
...
+};
+
+unique_ptr<FunctionData> ConditionCacheBuildBind(ClientContext &context, TableFunctionBindInput &input,
+                                                 vector<LogicalType> &return_types, vector<string> &names) {
+	auto result = make_uniq<ConditionCacheBuildBindData>();
+	result->table_name = input.inputs[0].GetValue<string>();
+	result->predicate_sql = input.inputs[1].GetValue<string>();
+
+	// Resolve table to get OID and total rows; throws if table doesn't exist
+	auto &table_entry = Catalog::GetEntry<DuckTableEntry>(context, INVALID_CATALOG, DEFAULT_SCHEMA, result->table_name);
```
dentiny: Same as my other comment, I think we need to handle fully/partial qualified table name, or and support entities.

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,152 @@
...
+	}
+	gstate.done = true;
+
+	const auto &bind_data = data_p.bind_data->Cast<ConditionCacheBuildBindData>();
+
+	// Run SELECT rowid on a new connection
+	string quoted_table = KeywordHelper::WriteQuoted(bind_data.table_name, '"');
+	string query = StringUtil::Format("SELECT rowid FROM %s WHERE %s", quoted_table, bind_data.predicate_sql);
+	auto &db = DatabaseInstance::GetDatabase(context);
+	Connection con(db);
```
dentiny: Is it possible to use duckdb internal interface here?

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,152 @@
...
+			row_group_vec_indices[row_group_idx].insert(vector_idx);
+			++total_qualifying_rows;
+		}
+	}
+
+	// Build cache entry from collected indices
+	auto entry = make_shared_ptr<ConditionCacheEntry>();
+	for (const auto &pair : row_group_vec_indices) {
+		vector<idx_t> indices(pair.second.begin(), pair.second.end());
+		entry->bitvectors.emplace(pair.first, RowGroupFilter(indices));
```
dentiny: We can take a span instead of a vector?

dentiny (commented):
some quick comments...

src/query_condition_cache_functions.cpp
dentiny: Two reasons
- You could have connection inside of connection -- I'm not sure how it works and if it works
- The query here will go through another full lifecycle of a complete query, including connection construction/destruction, query parsing/binding/execution/etc, which is the overhead I don't want to take

src/query_condition_cache_functions.cpp
Andrewtangtang: already in alphabetical order?

src/query_condition_cache_functions.cpp
Andrewtangtang: Currently only tables are supported since the cache operates at the row group / vector level of table storage. Curious what scenario would need caching on an index?

src/query_condition_cache_functions.cpp
Andrewtangtang: Ｉupdate a version uses internal APIs.

src/query_condition_cache_functions.cpp
Andrewtangtang: Curious about what is the advantage of `span` here?

test/sql/condition_cache_build.test
Andrewtangtang: I've opened an issue to track this #5  . Planning to extract filters from TableFilterSet, which is organized by column index, so the key would be canonical regardless of order.

src/query_condition_cache_functions.cpp
Andrewtangtang: but std::div is for signed types — idx_t is uint64_t.

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,234 @@
...
+	auto &transaction = DuckTransaction::Get(context, table_entry.ParentCatalog().GetAttached());
+	TableScanState scan_state;
+	storage.InitializeScan(context, transaction, scan_state, column_ids);
+
+	// Scan and evaluate predicate
+	DataChunk chunk;
+	chunk.Initialize(Allocator::Get(context), scan_types);
+	ExpressionExecutor executor(context, bound_expr);
+
+	unordered_map<idx_t, unordered_set<idx_t>> row_group_vec_indices;
```
peterxcli:
```suggestion
	unordered_map<idx_t, vector<idx_t>> row_group_vec_indices;
```
is the vector idx possible to be duplicated?

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,234 @@
...
+		chunk.Reset();
+		storage.Scan(transaction, chunk, scan_state);
+		if (chunk.size() == 0) {
+			break;
+		}
+
+		SelectionVector sel(chunk.size());
+		idx_t match_count = executor.SelectExpression(chunk, sel);
+
+		auto &rowid_vec = chunk.data[rowid_col_idx];
```
peterxcli, line 158: 1. what is the chunk data layout?
is it sth like this:
```
chunk 1:
rg, range of rg data
1, [80-400]
4, [12800-12880]
6, [1000-2000]

chunk 2:
rg, range of rg data
2, [90-100]
3, [10000-12280]
5, [1-2]
```
?

2. so it is possible that the same (row group idx, vector idx) pair will appear across multiple chunk?

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+	// Parse predicate SQL and bind column references to table columns
+	auto parsed_exprs = Parser::ParseExpressionList(bind_data.predicate_sql);
+	if (parsed_exprs.empty()) {
+		throw InvalidInputException("condition_cache_build: failed to parse predicate");
+	}
+	auto binder = Binder::CreateBinder(context);
+	physical_index_set_t bound_columns;
+	CheckBinder check_binder(*binder, context, bind_data.table_name, table_entry.GetColumns(), bound_columns);
+	auto bound_expr = check_binder.Bind(parsed_exprs[0]);
```
peterxcli: what is this for, sorry I dont have related knowledge.

src/query_condition_cache_functions.cpp
peterxcli: sorry the `rg` is `row_id`
`rg = row_id / RW_SIZE`
`vector_idx = (row_id % RW_SIZE) / VECTOR_SIZE`

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+	// SelectExpression requires BOOLEAN return type
+	if (bound_expr->return_type.id() != LogicalTypeId::BOOLEAN) {
+		bound_expr = BoundCastExpression::AddCastToType(context, std::move(bound_expr),
+		                                               LogicalType {LogicalTypeId::BOOLEAN});
+	}
```
peterxcli: what kind of the predicate will need this?

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,234 @@
...
+	idx_t total_row_groups = (bind_data.total_rows + DEFAULT_ROW_GROUP_SIZE - 1) / DEFAULT_ROW_GROUP_SIZE;
+
+	idx_t qualifying_vectors = 0;
+	for (const auto &bitvector_pair : entry->bitvectors) {
+		for (idx_t v = 0; v < vectors_per_row_group; ++v) {
+			if (bitvector_pair.second.VectorHasRows(v)) {
+				++qualifying_vectors;
+			}
+		}
+	}
```
peterxcli: I thnik this number can be calculate in `BuildCacheEntry`, too?

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+	RemapColumnIndices(bound_expr, storage_to_scan_idx);
+
+	// Direct table scan with predicate evaluation via ExpressionExecutor
+	auto &transaction = DuckTransaction::Get(context, table_entry.ParentCatalog().GetAttached());
+	TableScanState scan_state;
+	storage.InitializeScan(context, transaction, scan_state, column_ids);
+
+	// Scan and evaluate predicate
+	DataChunk chunk;
```
peterxcli: is this chunk materialised or not?

peterxcli: otherwise we are doing the full table scan here

src/query_condition_cache_functions.cpp
peterxcli: self answered. yes

src/query_condition_cache_functions.cpp
Andrewtangtang: Most predicates (val = 42, id > 10) already return BOOLEAN.
This handles edge cases like bare column references (e.g.
predicate = 'val') where the expression returns INTEGER.
SelectExpression requires BOOLEAN input.

peterxcli: why duckdb comparsion return int? is it intentional for specific reason?
from my intuition, `<`, `>` and `=` should all be logical output.

Andrewtangtang: CheckBinder::Bind returns a bound expression tree where
each node has a return_type. For 'val = 42':
                                                                                                  
```
BoundComparisonExpression(=)           return_type = BOOLEAN
├── BoundReferenceExpression(index=1)  return_type = INTEGER
└── BoundConstantExpression(42)        return_type = INTEGER
```                            

The root node's return_type is BOOLEAN, so no cast needed.
For a bare column like 'val':
```
BoundReferenceExpression(index=1)      return_type = INTEGER
```
The root return_type is INTEGER, so I wrap it with a
BoundCastExpression to BOOLEAN before passing to
SelectExpression which requires BOOLEAN input.

src/query_condition_cache_functions.cpp
Andrewtangtang: Yes, the chunk is materialized and this is a full table scan. I could add `TableFilterSet` to skip row groups via zone maps in a follow-up PR. BTW, I think the cache is most valuable for                             predicates that can't be pushed down (e.g. LIKE, OR), which would still require a full table scan to build.

peterxcli: > BTW, I think the cache is most valuable for predicates that can't be pushed down (e.g. LIKE, OR), which would still require a full table scan to build.

```
proj
^
|
Logical Filter (col2 LIKE "%1234%")
^
|
Logical Get (col1 < 1)
```

wont be full(materalized) table scan

src/query_condition_cache_functions.cpp
peterxcli: is it possible that user will build cache on a table the predicate "val"(one of the columns of that table)?

Andrewtangtang: I will change to throw an error.

src/query_condition_cache_functions.cpp
dentiny: my bad, must be me mis-reading :cry:

src/query_condition_cache_functions.cpp
dentiny: > Curious what scenario would need caching on an index?

Sorry I think it wrongly -- I was thinking if we could build a cache upon index scan, but it should be reflected to table rows as well.

dentiny: > Also, we should be able to handle <database>.<schema>.<object>, right?


Sure, the hierarchy here is used to locate a table.

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+namespace duckdb {
+namespace {
+
+// ------- condition_cache_build(table_name VARCHAR, predicate VARCHAR) -------
+
+struct ConditionCacheBuildBindData : public TableFunctionData {
+	string catalog;
+	string schema;
+	string table_name;
```
dentiny:
```suggestion
	string table;
```
just to keep naming consistency :)

src/include/query_condition_cache_functions.hpp
```diff
@@ -0,0 +1,20 @@
...
+
+namespace duckdb {
+
+class DuckTableEntry;
+class Expression;
+
+// Scan all rows in the table, evaluate bound_expr, and build a ConditionCacheEntry
+// recording which vectors contain qualifying rows.
+// Modifies bound_expr in-place (remaps column indices to scan positions).
+shared_ptr<ConditionCacheEntry> BuildCacheEntry(ClientContext &context, DuckTableEntry &table_entry,
```
dentiny, line 16: Wondering why do we expose this function? Do you plan to use the function in the followup PRs?

src/query_condition_cache_extension.cpp
```diff
@@ -1,14 +1,15 @@
 #define DUCKDB_EXTENSION_MAIN
 
 #include "query_condition_cache_extension.hpp"
+#include "query_condition_cache_functions.hpp"
```
dentiny: nit (we could discuss): it seems you want to split the inclusion section for duckdb headers and our own, but it seems to be the general coding practice is 
- corresponding header
- newliner
- others

test/sql/condition_cache_build.test
```diff
@@ -0,0 +1,51 @@
...
+
+# Invalid predicate throws error
+statement error
+SELECT * FROM condition_cache_build('t', 'invalid_col = 42');
+----
+does not contain column invalid_col
+
+# Qualified table name with schema
+query I
+SELECT status FROM condition_cache_build('main.t', 'val = 42');
```
dentiny, line 43: could you please also add test cases where
- database name is valid, schema name not valid
- database name not valid, but schema name valid

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+// Scan all rows, evaluate predicate, and build a ConditionCacheEntry.
+// Modifies bound_expr in-place (remaps column indices to scan positions).
+shared_ptr<ConditionCacheEntry> BuildCacheEntry(ClientContext &context, DuckTableEntry &table_entry,
+                                                Expression &bound_expr, idx_t &total_qualifying_rows) {
+	auto &storage = table_entry.GetStorage();
+	auto &columns = table_entry.GetColumns();
+
+	// Scan all physical columns (needed by the predicate expression) plus rowid
+	vector<StorageIndex> column_ids;
```
dentiny: from performance's perspective, we could store a struct of (storage index, logical type) in one vector

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,234 @@
...
+	// Scan and evaluate predicate
+	DataChunk chunk;
+	chunk.Initialize(Allocator::Get(context), scan_types);
+	ExpressionExecutor executor(context, bound_expr);
+
+	unordered_map<idx_t, unordered_set<idx_t>> row_group_vec_indices;
+	total_qualifying_rows = 0;
+
+	while (true) {
+		chunk.Reset();
```
dentiny: what about we define `DataChunk chunk;` inside of the loop?

src/include/query_condition_cache_functions.hpp
dentiny: One benefit to have it in the header file is maybe we could add a C++ unit test?
Just personal preference, I personally rely on these unit tests to understand the code logic.

dentiny (commented):
Thanks for the PR! I learnt a lot, left some trivial comments

src/query_condition_cache_functions.cpp
Andrewtangtang: Sorry for previous inaccurate statement. CheckBinder was originally designed for CHECK constraints (validating INSERT/UPDATE data), so it sets `target_type = INTEGER` and wraps every bound expression in a cast to INTEGER. The actual tree for `val = 42` is:

```
BoundCastExpression(INTEGER)              ← return_type = INTEGER (added by CheckBinder)
└── BoundComparisonExpression(=)          ← return_type = BOOLEAN
    ├── BoundReferenceExpression(index=1) ← return_type = INTEGER
    └── BoundConstantExpression(42)       ← return_type = INTEGER
```

The root return_type is INTEGER, not BOOLEAN. So the cast back to BOOLEAN is always needed.

For bare column references like `'val'`, I now reject them before binding — if the parsed expression is a `ColumnRefExpression`, I throw an error: `"predicate must be a boolean expression, not a column reference"`.

src/include/query_condition_cache_functions.hpp
Andrewtangtang: I am planning to use in the future and I add a C++ unit test accordingly.

src/query_condition_cache_functions.cpp
Andrewtangtang: Each chunk from `DataTable::Scan` returns `STANDARD_VECTOR_SIZE` contiguous rows      
sequentially, so one chunk maps to exactly one (row_group_idx, vector_idx) pair. 
The same pair won't appear across multiple chunks.

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+// ------- condition_cache_build(table_name VARCHAR, predicate VARCHAR) -------
+
+struct ConditionCacheBuildBindData : public TableFunctionData {
+	string catalog;
+	string schema;
+	string table;
+	string predicate_sql;
+	idx_t table_oid;
+	idx_t total_rows;
```
dentiny: Do you mind adding a comment does the row # mean "total number of rows in the table" or "number of rows matching the predicate"?

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+unique_ptr<FunctionData> ConditionCacheBuildBind(ClientContext &context, TableFunctionBindInput &input,
+                                                 vector<LogicalType> &return_types, vector<string> &names) {
+	auto result = make_uniq<ConditionCacheBuildBindData>();
+	result->predicate_sql = input.inputs[1].GetValue<string>();
+
+	// Parse qualified table name (supports "table", "schema.table", "catalog.schema.table")
+	auto qname = QualifiedName::Parse(input.inputs[0].GetValue<string>());
+	Binder::BindSchemaOrCatalog(context, qname.catalog, qname.schema);
+	result->catalog = qname.catalog;
```
dentiny:
```suggestion
	result->catalog = std::move(qname.catalog);
```

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+	idx_t scan_pos = 0;
+	for (auto &col : columns.Physical()) {
+		scan_columns.emplace_back(StorageIndex(col.Oid()), col.Type());
+		storage_to_scan_idx[col.Oid()] = scan_pos++;
+	}
+	const idx_t rowid_col_idx = scan_pos;
+	scan_columns.emplace_back(StorageIndex(COLUMN_IDENTIFIER_ROW_ID), LogicalType::ROW_TYPE);
+
+	vector<StorageIndex> column_ids;
```
dentiny, line 126: BTW generally I prefer struct over pair, since 
- All fields are named, which is better for readability
- Easier to extend if we want to add additional fields

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,246 @@
...
+                                                Expression &bound_expr) {
+	auto &storage = table_entry.GetStorage();
+	auto &columns = table_entry.GetColumns();
+
+	// Scan all physical columns (needed by the predicate expression) plus rowid
+	vector<pair<StorageIndex, LogicalType>> scan_columns;
+	unordered_map<idx_t, idx_t> storage_to_scan_idx;
+
+	idx_t scan_pos = 0;
+	for (auto &col : columns.Physical()) {
```
dentiny:
```suggestion
	for (const auto &col : columns.Physical()) {
```
is there any reason we need to mutate?

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,246 @@
...
+// Scan all rows, evaluate predicate, and build a ConditionCacheEntry.
+// Modifies bound_expr in-place (remaps column indices to scan positions).
+shared_ptr<ConditionCacheEntry> BuildCacheEntry(ClientContext &context, DuckTableEntry &table_entry,
+                                                Expression &bound_expr) {
+	auto &storage = table_entry.GetStorage();
+	auto &columns = table_entry.GetColumns();
+
+	// Scan all physical columns (needed by the predicate expression) plus rowid
+	vector<pair<StorageIndex, LogicalType>> scan_columns;
+	unordered_map<idx_t, idx_t> storage_to_scan_idx;
```
dentiny: I would suggest to leave some comments: it's mapped from column OID to scan index.

dentiny:
```suggestion
	unordered_map<column_t, idx_t> storage_to_scan_idx;
```

dentiny: which is the type value for `Oid`
https://github.com/duckdb/duckdb/blob/dc11eadd8f0a7c600f0034810706605ebe10d5b9/src/include/duckdb/parser/column_definition.hpp#L62

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+// Scan all rows, evaluate predicate, and build a ConditionCacheEntry.
+// Modifies bound_expr in-place (remaps column indices to scan positions).
+shared_ptr<ConditionCacheEntry> BuildCacheEntry(ClientContext &context, DuckTableEntry &table_entry,
+                                                Expression &bound_expr) {
+	auto &storage = table_entry.GetStorage();
+	auto &columns = table_entry.GetColumns();
+
+	// Scan all physical columns (needed by the predicate expression) plus rowid
+	vector<pair<StorageIndex, LogicalType>> scan_columns;
```
dentiny: Both containers should resize with `columns.LogicalColumnCount()`

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,246 @@
...
+	unordered_map<idx_t, idx_t> storage_to_scan_idx;
+
+	idx_t scan_pos = 0;
+	for (auto &col : columns.Physical()) {
+		scan_columns.emplace_back(StorageIndex(col.Oid()), col.Type());
+		storage_to_scan_idx[col.Oid()] = scan_pos++;
+	}
+	const idx_t rowid_col_idx = scan_pos;
+	scan_columns.emplace_back(StorageIndex(COLUMN_IDENTIFIER_ROW_ID), LogicalType::ROW_TYPE);
```
dentiny:
```suggestion
	scan_columns.emplace_back(StorageIndex(COLUMN_IDENTIFIER_ROW_ID), LogicalType {LogicalTypeId::ROW_TYPE});
```

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+	vector<StorageIndex> column_ids;
+	vector<LogicalType> scan_types;
+	for (const auto &[id, type] : scan_columns) {
+		column_ids.push_back(id);
```
dentiny: I don't understand why we cannot construct at L99?

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,246 @@
...
+	return make_uniq<ConditionCacheBuildState>();
+}
+
+// After binding, column indices refer to StorageOid which may not match the scan order.
+// Remap them to the actual DataChunk column positions.
+void RemapColumnIndices(Expression &expr, const unordered_map<idx_t, idx_t> &mapping) {
+	if (expr.GetExpressionClass() == ExpressionClass::BOUND_REF) {
+		auto &ref = expr.Cast<BoundReferenceExpression>();
+		auto it = mapping.find(ref.index);
+		if (it != mapping.end()) {
```
dentiny: curious in which case expr index doesn't exist in mapping?

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,246 @@
...
+	auto &storage = table_entry.GetStorage();
+	auto &columns = table_entry.GetColumns();
+
+	// Scan all physical columns (needed by the predicate expression) plus rowid
+	vector<pair<StorageIndex, LogicalType>> scan_columns;
+	unordered_map<idx_t, idx_t> storage_to_scan_idx;
+
+	idx_t scan_pos = 0;
+	for (auto &col : columns.Physical()) {
+		scan_columns.emplace_back(StorageIndex(col.Oid()), col.Type());
```
dentiny: I think I might be wrong at https://github.com/dentiny/duckdb-query-condition-cache/pull/6#discussion_r2889312965, these two vectors are needed separately, so have to get two vectors

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+	// Direct table scan with predicate evaluation via ExpressionExecutor
+	auto &transaction = DuckTransaction::Get(context, table_entry.ParentCatalog().GetAttached());
+	TableScanState scan_state;
+	storage.InitializeScan(context, transaction, scan_state, column_ids);
+
+	// Scan and evaluate predicate
+	ExpressionExecutor executor(context, bound_expr);
+
+	unordered_map<idx_t, unordered_set<idx_t>> row_group_vec_indices;
```
dentiny: nit: add a comment
```
// Maps from xxx to yyy.
```

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,246 @@
...
+		chunk.Initialize(Allocator::Get(context), scan_types);
+		storage.Scan(transaction, chunk, scan_state);
+		if (chunk.size() == 0) {
+			break;
+		}
+
+		SelectionVector sel(chunk.size());
+		idx_t match_count = executor.SelectExpression(chunk, sel);
+
+		auto &rowid_vec = chunk.data[rowid_col_idx];
```
dentiny, line 158: IIUC, our current implementation returns all columns in the data chunk, but we only need row id set our own filter bit; I'm wondering if it's return rowid only?

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,246 @@
...
+		UnifiedVectorFormat rowid_data;
+		rowid_vec.ToUnifiedFormat(chunk.size(), rowid_data);
```
dentiny, line 160: IIUC, our current implementation returns all columns in the data chunk, but we only need row id set our own filter bit; I'm wondering if it's return rowid only (could be in a followup PR, or I could take a look)?

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,246 @@
...
+		auto rowids = UnifiedVectorFormat::GetData<row_t>(rowid_data);
+
+		for (idx_t i = 0; i < match_count; ++i) {
```
dentiny: BTW, I usually use `idx` or `ii` as index, the reason being more grep-friendly

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,246 @@
...
+			auto rid_idx = rowid_data.sel->get_index(sel.get_index(i));
```
dentiny: worth giving a better name

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,246 @@
...
+			if (!rowid_data.validity.RowIsValid(rid_idx)) {
+				continue;
+			}
+			auto row_id = NumericCast<idx_t>(rowids[rid_idx]);
+			idx_t row_group_idx = row_id / DEFAULT_ROW_GROUP_SIZE;
+			idx_t vector_idx = (row_id % DEFAULT_ROW_GROUP_SIZE) / STANDARD_VECTOR_SIZE;
+			row_group_vec_indices[row_group_idx].insert(vector_idx);
```
dentiny: Can we use bitmap to store vector indexes? current implementation = insert into hashset + to vector + bitmap

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,246 @@
...
+	idx_t remaining_rows = total_rows % DEFAULT_ROW_GROUP_SIZE;
+	idx_t total_vectors = full_row_groups * vectors_per_row_group;
+	if (remaining_rows > 0) {
+		total_vectors += (remaining_rows + STANDARD_VECTOR_SIZE - 1) / STANDARD_VECTOR_SIZE;
+	}
+
+	idx_t qualifying_row_groups = entry.bitvectors.size();
+	idx_t total_row_groups = (total_rows + DEFAULT_ROW_GROUP_SIZE - 1) / DEFAULT_ROW_GROUP_SIZE;
+
+	return {qualifying_vectors, total_vectors, qualifying_row_groups, total_row_groups};
```
dentiny: Can we do 
```cpp
CacheEntryStats {
 .a = a,
};
```
The reason being, all data fields are of the same type, it's easy to mis-order; and if we reorder field type in the struct, compilation still passes.

test/unittest/test_build_cache_entry.cpp
```diff
@@ -0,0 +1,79 @@
...
+			                                               LogicalType {LogicalTypeId::BOOLEAN});
+		}
+
+		auto entry = BuildCacheEntry(context, table_entry, *bound_expr);
+
+		REQUIRE(entry != nullptr);
+		REQUIRE(entry->bitvectors.size() == 5); // 5 row groups
+	}
+
+	SECTION("selective predicate") {
```
dentiny, line 43: please add a unit test where odd value passes while even values don't

test/unittest/test_build_cache_entry.cpp
```diff
@@ -0,0 +1,79 @@
...
+		auto parsed_exprs = Parser::ParseExpressionList("id < 3000");
+		auto binder = Binder::CreateBinder(context);
+		physical_index_set_t bound_columns;
+		CheckBinder check_binder(*binder, context, "t", table_entry.GetColumns(), bound_columns);
+		auto bound_expr = check_binder.Bind(parsed_exprs[0]);
```
dentiny, line 48: Does check binder always return int?
https://github.com/duckdb/duckdb/blob/fbeb7936352b2d32ad8562f19eab8fdfcf430e61/src/planner/expression_binder/check_binder.cpp#L12
so we don't need the if statement here?

dentiny:
Ping me when ready, thanks!

src/query_condition_cache_functions.cpp
Andrewtangtang: It shouldn't happen. Changed if to `D_ASSERT`.

src/query_condition_cache_functions.cpp
Andrewtangtang: We can't scan rowid only. `ExpressionExecutor::SelectExpression` needs the column data to evaluate the predicate. But I changed to scan only the necessary columns + rowid now.

src/query_condition_cache_functions.cpp
Andrewtangtang: Designated initializers (.field = value) trigger a -Wc++20-designator compiler warning. So I used named assignment instead.

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,264 @@
...
+		auto rowids = UnifiedVectorFormat::GetData<row_t>(rowid_data);
+
+		for (idx_t idx = 0; idx < match_count; ++idx) {
+			auto rowid_entry = rowid_data.sel->get_index(sel.get_index(idx));
+			if (!rowid_data.validity.RowIsValid(rowid_entry)) {
+				continue;
+			}
+			auto row_id = NumericCast<idx_t>(rowids[rowid_entry]);
+			idx_t row_group_idx = row_id / DEFAULT_ROW_GROUP_SIZE;
+			idx_t vector_idx = (row_id % DEFAULT_ROW_GROUP_SIZE) / STANDARD_VECTOR_SIZE;
```
peterxcli, line 170: nit: extract the logic and reuse

sth like
```suggestion
			auto [row_group_idx, vector_idx] = ToXXX(row_id);
```
can be followup

peterxcli (approved):
lgtm. thanks for all the updates and response.

Andrewtangtang:
@dentiny Please take another look. Thanks!

src/query_condition_cache_functions.cpp
dentiny: Yeah I'm aware of it, I think we could suppress the warning (see my cache httpfa extension setup) , but no string opinion, up to you

Andrewtangtang: Done.

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,264 @@
...
+	                                      [&](Expression &child) { CollectReferencedColumns(child, column_oids); });
+}
+
+// After binding, column indices refer to StorageOid which may not match the scan order.
+// Remap them to the actual DataChunk column positions.
+void RemapColumnIndices(Expression &expr, const unordered_map<column_t, idx_t> &mapping) {
+	if (expr.GetExpressionClass() == ExpressionClass::BOUND_REF) {
+		auto &ref = expr.Cast<BoundReferenceExpression>();
+		auto it = mapping.find(ref.index);
+		D_ASSERT(it != mapping.end());
```
dentiny, line 85: FYI, in DuckDB v1.5, we could use https://github.com/duckdb/duckdb/blob/3d63e1d1197a6f7413da25d040ab7b5fdebdbdaf/src/include/duckdb/common/assert.hpp#L43

src/query_condition_cache_functions.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+	// Parse predicate SQL and bind column references to table columns
+	auto parsed_exprs = Parser::ParseExpressionList(bind_data.predicate_sql);
+	if (parsed_exprs.empty()) {
+		throw InvalidInputException("condition_cache_build: failed to parse predicate");
+	}
+	// Reject bare column references (e.g. predicate = 'val')
+	if (parsed_exprs[0]->GetExpressionClass() == ExpressionClass::COLUMN_REF) {
+		throw InvalidInputException(
+		    "condition_cache_build: predicate must be a boolean expression, not a column reference");
```
dentiny: Can we use DuckDB internal functions to detect predicate return value? I feel like we need to check a bunch of possibilities.

Andrewtangtang: I removed it to be consistent with DuckDB SQL semantics. The `CheckBinder` and `Parser` already have their own mechanism to handle invalid predicates           internally. I added tests to verify these error cases.

dentiny (approved):
Thanks!

PR #16, opened 2026-03-11
[Demo] Automatic cache build

### LINKED ISSUE #11: [Demo] Complete build cache automatically when running the query
Asking users to extract their predicate and use it as extension scalar function parameter is not intuitive. Some transparent or straightforward way must provide.

peterxcli:
Closes: https://github.com/dentiny/duckdb-query-condition-cache/issues/11

## Summary

This PR adds automatic query condition cache build and apply via a two-phase optimizer extension. When `use_query_condition_cache` is enabled, the pre-optimize phase captures the full WHERE clause before FilterPushdown, computes a canonical predicate key, and builds the cache inline on miss so the first query benefits immediately. The post-optimize phase injects a `CacheExpressionFilter` on ROW_ID into `LogicalGet.table_filters`, providing row-group pruning via `CheckStatistics` and vector-level filtering via bitvector lookup. Benchmarks are included demonstrating significant speedups for non-pushable predicates (OR, compound LIKE OR) where DuckDB otherwise must decompress all columns for all vectors.

## Example Usage

```sql
-- Enable automatic cache
SET use_query_condition_cache = true;

-- First query: builds cache inline during planning, then executes with cache filter
SELECT sum(length(col1)), sum(length(col2)) FROM events WHERE text_a LIKE '%ERROR%' OR text_b LIKE '%WARN%';

-- Second query: cache hit, skips vectors with no qualifying rows
SELECT sum(length(col1)), sum(length(col2)) FROM events WHERE text_a LIKE '%ERROR%' OR text_b LIKE '%WARN%';

-- Disable when not needed
SET use_query_condition_cache = false;
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,9 +15,12 @@ endif()
 
 include_directories(src/include)
 
-set(EXTENSION_SOURCES src/query_condition_cache_extension.cpp
-                      src/query_condition_cache_functions.cpp
-                      src/query_condition_cache_state.cpp)
+set(EXTENSION_SOURCES
+    src/query_condition_cache_extension.cpp
+    src/query_condition_cache_filter.cpp
+    src/query_condition_cache_functions.cpp
+    src/query_condition_cache_optimizer.cpp
+    src/query_condition_cache_state.cpp)
 
 build_static_extension(${TARGET_NAME} ${EXTENSION_SOURCES})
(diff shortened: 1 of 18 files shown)


peterxcli:
already merged in https://github.com/Andrewtangtang/duckdb-query-condition-cache/pull/1

PR #20, opened 2026-03-12
Use object cache LRU cache for predicate cache entries

### LINKED ISSUE #7: Implement eviction for query condition cache entry
Query condition cache cannot grow indefinitely, I added a per-database LRU cache in DuckDB v1.5, which will be released next week.
I will take care of the eviction and integration.

dentiny:
Closes: https://github.com/dentiny/duckdb-query-condition-cache/issues/7

This PR integrates predicate cache entries into object cache, which is a DuckDB internal LRU cache.
The benefit for the integration is it's accounted into whole database instance's resource consumption, and is managed by `max_memory` config.
I removed two functions, which are definitely not supported for now; let me know if they're really needed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,9 +15,9 @@ endif()
 
 include_directories(src/include)
 
-set(EXTENSION_SOURCES src/query_condition_cache_extension.cpp
-                      src/query_condition_cache_functions.cpp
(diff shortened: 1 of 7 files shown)


peterxcli (approved):
Thanks! builtin object store lru cache is clean and great.

PR #26, opened 2026-03-12
Add TPCH Benchmark

### LINKED ISSUE #25: [Demo] TPCH and TPCDH Benchmark
as title. 

you might want to use the duckdb existing slt TPCH and TPCDH suite to do this.

peterxcli:
Closes: https://github.com/dentiny/duckdb-query-condition-cache/issues/25

Per-query protocol (all runs in the same DuckDB session):
-   Run 1 – cache OFF, OS cold  : cold baseline (reference only)
-   Run 2 – cache OFF, OS warm  : true baseline (warm OS, no condition cache)
-   Run 3 – cache ON,  OS warm  : cache build (first query with cache enabled)
-   Run 4 – cache ON,  OS warm  : cache hit   (warm OS + warm condition cache)

  Speedup = Run 2 / Run 4

Runs 2 and 4 are each repeated REPEAT times; the median is used.


peterxcli:
image

PR #29, opened 2026-03-12
[Demo] parallel cache build

Andrewtangtang:


src/query_condition_cache_functions.cpp
```diff
@@ -93,9 +96,76 @@ struct ScanColumn {
...
+	void ExecuteTask() override {
+		TableScanState scan_state;
+		scan_state.Initialize(column_ids);
+
+		ExpressionExecutor expr_executor(context, bound_expr);
+
+		DataChunk chunk;
+		chunk.Initialize(Allocator::Get(context), scan_types);
+
+		while (storage.NextParallelScan(context, parallel_state, scan_state)) {
```
dentiny, line 120: Curious how much is the performance change? Also wondering if we need to parallelize the internal while true block?

dentiny:
If you suspect it's a caused by optimizations to prune table scan, we could maybe revert back to use SQL.

Andrewtangtang:
I tested the current version by running the TPC-H benchmark (SF=1 and SF=10). The build
time with the parallel DataTable::Scan approach is approximately the same as the SQL-based approach, so I think we can keep this implementation.

peterxcli (approved):

PR #35, opened 2026-03-15
[Demo] perf: optimizer early return and filter always true

peterxcli:
This pull request improves the efficiency and correctness of the query condition cache and its integration with DuckDB's filter pushdown logic. The main changes include adding a check for fully-matching row group filters, optimizing when the cache is used by skipping cases already handled natively by DuckDB, and refactoring to support these enhancements.

**Improvements to filter pruning logic:**

* Added a new `IsFull()` method to the `RowGroupFilter` class, allowing detection when all vectors in a row group match, and implemented its logic in `src/query_condition_cache_state.cpp`. [[1]](diffhunk://#diff-88a03539ebda86ce9381b8dd9b5d0a8fa9639b5a97acb89191dbe91175a0e53bR33) [[2]](diffhunk://#diff-dafe1beabdcd8b768bb9fd3f2c978e072c365c11ae84dfd4b65e858b4362fca1R34-R50)
* Updated the filter pruning decision in `CacheExpressionFilter::CheckStatistics` to return `FILTER_ALWAYS_TRUE` if a row group filter is fully matching, and refactored the logic for clarity and correctness.

**Optimization of cache usage:**

* Introduced the `IsTriviallyPushable` function to detect filter expressions that DuckDB's native optimizer already handles efficiently (e.g., comparisons with constants, BETWEEN, IS NULL/NOT NULL), and skip cache processing for these cases in `QueryConditionCacheOptimizer::ProcessFilters`. [[1]](diffhunk://#diff-0160a3c7dde6d7f85717e21ff8cb39bbcf3f68e0ea8385d540d43c2fb86a4446R81-R101) [[2]](diffhunk://#diff-0160a3c7dde6d7f85717e21ff8cb39bbcf3f68e0ea8385d540d43c2fb86a4446R139-R152)
* Added necessary includes for new expression types used in the pushdown detection logic.

### DIFF
--- a/src/query_condition_cache_optimizer.cpp
+++ b/src/query_condition_cache_optimizer.cpp
@@ -15,6 +15,8 @@
 #include "duckdb/planner/expression/bound_between_expression.hpp"
 #include "duckdb/planner/expression/bound_columnref_expression.hpp"
 #include "duckdb/planner/expression/bound_comparison_expression.hpp"
+#include "duckdb/planner/expression/bound_conjunction_expression.hpp"
+#include "duckdb/planner/expression/bound_constant_expression.hpp"
 #include "duckdb/planner/expression/bound_operator_expression.hpp"
 #include "duckdb/planner/expression_iterator.hpp"
 #include "duckdb/planner/expression_binder/check_binder.hpp"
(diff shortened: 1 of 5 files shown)


src/query_condition_cache_optimizer.cpp
```diff
@@ -75,6 +78,27 @@ void QueryConditionCacheOptimizer::CollectGets(unique_ptr<LogicalOperator> &plan
 	}
 }
 
+// Check if an expression will be natively pushed down by DuckDB's FilterPushdown optimizer,
+// meaning the condition cache provides no additional benefit.
+// Covers: col op constant, BETWEEN with constant bounds, IS NULL, IS NOT NULL.
+static bool IsTriviallyPushable(const Expression &expr) {
```
peterxcli, line 86: let me review this first.

PR #37, opened 2026-03-17
Replace thread_local with ClientContext::registered_state for optimiz…

Andrewtangtang:
This PR replaces the `thread_local` map used to pass `ConditionCacheEntry` data between the pre-optimize and post-optimize phases with a `CacheOptimizerQueryState` stored in `ClientContext::registered_state`. This ties the optimizer state to the query lifecycle and gets automatic `QueryEnd` cleanup


peterxcli (approved):
Thanks!

PR #41, opened 2026-03-22
Parallelize cache build and record all row groups

Andrewtangtang:
## Summary

This PR parallelizes `BuildCacheEntry` using `TaskExecutor` with per-thread local entries merged via a new `RowGroupFilter::MergeFrom`, and records all scanned row groups in cache entries including those with no qualifying rows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,3 +29,7 @@ install(
   ARCHIVE DESTINATION "${INSTALL_LIB_DIR}")
 
 add_subdirectory(test/unittest)
+
+add_executable(benchmark_cache_build EXCLUDE_FROM_ALL
+               benchmark/benchmark_cache_build.cpp)
+target_link_libraries(benchmark_cache_build duckdb ${EXTENSION_NAME})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,7 +30,7 @@ install(
 
 add_subdirectory(test/unittest)
 
-if(BUILD_BENCHMARK)
+if(${BUILD_BENCHMARKS})
   add_executable(benchmark_cache_build benchmark/benchmark_cache_build.cpp)
   target_link_libraries(benchmark_cache_build duckdb ${EXTENSION_NAME})
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,7 @@ install(
 
 add_subdirectory(test/unittest)
 
-add_executable(benchmark_cache_build EXCLUDE_FROM_ALL
-               benchmark/benchmark_cache_build.cpp)
-target_link_libraries(benchmark_cache_build duckdb ${EXTENSION_NAME})
+if(BUILD_BENCHMARK)
+  add_executable(benchmark_cache_build benchmark/benchmark_cache_build.cpp)
+  target_link_libraries(benchmark_cache_build duckdb ${EXTENSION_NAME})
+endif()
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ format-all: format
 	cmake-format -i test/unittest/CMakeLists.txt
 
 benchmark_cache_build:
-	cmake -DBUILD_BENCHMARK=ON -B build/release -S duckdb
+	cmake -DBUILD_BENCHMARKS=ON -B build/release -S duckdb
 	cmake --build build/release --target benchmark_cache_build
 	./build/release/extension/query_condition_cache/benchmark_cache_build
 
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,7 @@ format-all: format
 	cmake-format -i test/unittest/CMakeLists.txt
 
 benchmark_cache_build:
+	cmake -DBUILD_BENCHMARK=ON -B build/release -S duckdb
 	cmake --build build/release --target benchmark_cache_build
 	./build/release/extension/query_condition_cache/benchmark_cache_build
 
--- a/Makefile
+++ b/Makefile
@@ -30,4 +30,8 @@ format-all: format
 	cmake-format -i CMakeLists.txt
 	cmake-format -i test/unittest/CMakeLists.txt
 
-PHONY: format-all test_release_all test_debug_all test_reldebug_all
+benchmark_cache_build:
+	cmake --build build/release --target benchmark_cache_build
+	./build/release/extension/query_condition_cache/benchmark_cache_build
+
+PHONY: format-all test_release_all test_debug_all test_reldebug_all benchmark_cache_build
--- a/benchmark/benchmark_cache_build.cpp
+++ b/benchmark/benchmark_cache_build.cpp
@@ -1,0 +1,106 @@
+#include "duckdb/main/connection.hpp"
+#include "duckdb/main/database.hpp"
+
+#include <chrono>
+#include <cstdio>
+#include <string>
+#include <vector>
+
+using namespace duckdb;
+
+static double BenchMs(Connection &con, const std::string &sql, int rounds = 5) {
+	con.Query(sql); // warmup
+	double total = 0;
+	for (int r = 0; r < rounds; ++r) {
+		auto start = std::chrono::steady_clock::now();
+		con.Query(sql);
+		auto end = std::chrono::steady_clock::now();
+		total += std::chrono::duration<double, std::milli>(end - start).count();
+	}
+	return total / rounds;
+}
+
+static void PrintSection(const char *title, const std::vector<std::pair<std::string, double>> &rows) {
+	printf("\n%s\n", title);
+	for (auto &[label, ms] : rows) {
+		printf("  %-45s %8.1f ms\n", label.c_str(), ms);
+	}
+}
+
+int main() {
+	DuckDB db;
+	Connection con(db);
+
+	const int rounds = 5;
+	const std::string build = "SELECT * FROM condition_cache_build('t', 'val = 42')";
+
+	// 0. Query baseline vs cache build cost
+	con.Query("CREATE OR REPLACE TABLE t AS SELECT i AS id, i%100 AS val FROM range(2000000) t(i)");
+	std::vector<std::pair<std::string, double>> results;
+	results.emplace_back("SELECT count(*) WHERE val = 42",
+	                     BenchMs(con, "SELECT count(*) FROM t WHERE val = 42", rounds));
+	results.emplace_back("SELECT count(*) WHERE val < 50",
+	                     BenchMs(con, "SELECT count(*) FROM t WHERE val < 50", rounds));
+	results.emplace_back("cache build (val = 42)", BenchMs(con, build, rounds));
+	results.emplace_back("cache build (val < 50)",
+	                     BenchMs(con, "SELECT * FROM condition_cache_build('t', 'val < 50')", rounds));
+	PrintSection("Query baseline vs cache build cost (2M rows, threads=default)", results);
+
+	// 1. Row count scaling
+	results.clear();
+	for (int64_t n : {100000, 500000, 2000000, 12288000}) {
+		con.Query("CREATE OR REPLACE TABLE t AS SELECT i AS id, i%100 AS val FROM range(" + std::to_string(n) + ") t(i)");
+		double ms = BenchMs(con, build, rounds);
+		int rgs = (n + 122879) / 122880;
+		char label[64];
+		snprintf(label, sizeof(label), "%'lld rows (%d RGs)", (long long)n, rgs);
+		results.emplace_back(label, ms);
+	}
+	PrintSection("Row count scaling (threads=default)", results);
+
+	// 2. Thread scaling
+	con.Query("CREATE OR REPLACE TABLE t AS SELECT i AS id, i%100 AS val FROM range(2000000) t(i)");
+	results.clear();
+	for (int threads : {1, 2, 4, 8}) {
+		con.Query("SET threads = " + std::to_string(threads));
+		double ms = BenchMs(con, build, rounds);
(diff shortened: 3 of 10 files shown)


src/query_condition_cache_functions.cpp
```diff
@@ -93,9 +96,87 @@ struct ScanColumn {
...
 
+// Task that scans a subset of row groups in parallel and builds a local ConditionCacheEntry.
+class CacheBuildTask : public BaseExecutorTask {
+public:
+	CacheBuildTask(TaskExecutor &executor, ClientContext &context, DataTable &storage, DuckTransaction &transaction,
+	               ParallelTableScanState &parallel_state, Expression &bound_expr,
+	               const vector<StorageIndex> &column_ids, const vector<LogicalType> &scan_types, idx_t rowid_col_idx,
+	               ConditionCacheEntry &local_entry)
+	    : BaseExecutorTask(executor), context(context), storage(storage), transaction(transaction),
+	      parallel_state(parallel_state), bound_expr(bound_expr), column_ids(column_ids), scan_types(scan_types),
```
dentiny:
```suggestion
	      parallel_state(parallel_state_p), bound_expr(bound_expr_p), column_ids(column_ids_p), scan_types(scan_types_p),
```
DuckDB uses `_p` prefix to differentiate function arguments

src/query_condition_cache_functions.cpp
```diff
@@ -134,41 +215,38 @@ shared_ptr<ConditionCacheEntry> BuildCacheEntry(ClientContext &context, DuckTabl
...
+	storage.InitializeParallelScan(context, parallel_state, column_indexes);
 
-	while (true) {
-		DataChunk chunk;
-		chunk.Initialize(Allocator::Get(context), scan_types);
-		storage.Scan(transaction, chunk, scan_state);
-		if (chunk.size() == 0) {
-			break;
-		}
+	auto num_threads = NumericCast<idx_t>(TaskScheduler::GetScheduler(context).NumberOfThreads());
```
dentiny, line 235: curious what are the difference between ?
- TaskScheduler::GetScheduler(context).NumberOfThreads()
- storage.MaxThreads(context)

test/sql/condition_cache_build.test
```diff
@@ -54,7 +54,7 @@ Cache Built: 245/245 vectors, 5/5 row groups
 statement error
 SELECT * FROM condition_cache_build('bad_schema.t', 'val = 42');
 ----
-Table with name t does not exist
+does not exist
```
dentiny: Why this change?

dentiny (commented):
I'm thinking if we could have a benchmark file to check the improvement.
For example, certain questions could be answered:
- What's the rough construction time for query cache (baseline)?
- How does parallel scan help here?
- How does it change as we increase column number or row number or row group number?

Andrewtangtang:
I add a cache build benchmark.
```
make benchmark_cache_build
Query baseline vs cache build cost (2M rows, threads=default)
  SELECT count(*) WHERE val = 42                     0.4 ms
  SELECT count(*) WHERE val < 50                     0.3 ms
  cache build (val = 42)                             0.7 ms
  cache build (val < 50)                             1.3 ms

Row count scaling (threads=default)
  100000 rows (1 RGs)                                0.2 ms
  500000 rows (5 RGs)                                0.3 ms
  2000000 rows (17 RGs)                              0.5 ms
  12288000 rows (100 RGs)                            2.4 ms

Thread scaling (2M rows, 17 RGs)
  threads = 1                                        1.6 ms
  threads = 2                                        1.0 ms
  threads = 4                                        0.6 ms
  threads = 8                                        0.5 ms

Column count in predicate (2M rows, threads=default)
  1 col                                              0.5 ms
  3 cols                                             1.0 ms
  5 cols                                             1.5 ms

Selectivity (2M rows, threads=default)
  1% match (val = 42)                                0.5 ms
  50% match (val < 50)                               1.1 ms
  100% match (val >= 0)                              2.0 ms
  0% match (id < 0)                                  0.4 ms
```

src/query_condition_cache_functions.cpp
Andrewtangtang: [`storage.MaxThreads(context)`](https://github.com/duckdb/duckdb/blob/7e7ee2bdd7be363c44e8943c30c25386e0acab1d/src/storage/data_table.cpp#L267) returns the row group count, and `TaskScheduler::GetScheduler(context).NumberOfThreads()` returns the thread pool size in DuckDB, which can be set via `SET threads = N`.

CMakeLists.txt
```diff
@@ -29,3 +29,7 @@ install(
   ARCHIVE DESTINATION "${INSTALL_LIB_DIR}")
 
 add_subdirectory(test/unittest)
+
+add_executable(benchmark_cache_build EXCLUDE_FROM_ALL
```
dentiny: benchmark is by default not built for duckdb, let's add a if
https://github.com/dentiny/duck-read-cache-fs/blob/2732962e4ab150f12a1d34eb164191bfa6f1a6b3/CMakeLists.txt#L146

benchmark/benchmark_cache_build.cpp
```diff
@@ -0,0 +1,106 @@
+#include "duckdb/main/connection.hpp"
+#include "duckdb/main/database.hpp"
+
+#include <chrono>
+#include <cstdio>
+#include <string>
```
dentiny: both string and vector should use duckdb headers, which adds some access safeguards

benchmark/benchmark_cache_build.cpp
```diff
@@ -0,0 +1,106 @@
...
+#include <vector>
+
+using namespace duckdb;
+
+static double BenchMs(Connection &con, const std::string &sql, int rounds = 5) {
```
dentiny: nit: prefer anonymous namespace

src/query_condition_cache_functions.cpp
```diff
@@ -93,9 +96,88 @@ struct ScanColumn {
...
+				// Determine row group from first row_id and instantiate its entry
+				// so that fully excluded row groups are recorded with empty bitvectors.
+				auto first_rowid_idx = rowid_data.sel->get_index(0);
+				if (!rowid_data.validity.RowIsValid(first_rowid_idx)) {
+					continue;
+				}
+				auto first_row_id = NumericCast<idx_t>(rowids[first_rowid_idx]);
+				idx_t row_group_idx = first_row_id / DEFAULT_ROW_GROUP_SIZE;
+				// Access to instantiate default RowGroupFilter for this row group
+				(void)local_entry.bitvectors[row_group_idx];
```
dentiny, line 146: not sure why it's needed?

src/query_condition_cache_functions.cpp
```diff
@@ -134,41 +216,38 @@ shared_ptr<ConditionCacheEntry> BuildCacheEntry(ClientContext &context, DuckTabl
...
-		SelectionVector sel(chunk.size());
-		idx_t match_count = executor.SelectExpression(chunk, sel);
+	vector<ConditionCacheEntry> local_entries(num_tasks);
 
-		auto &rowid_vec = chunk.data[rowid_col_idx];
-		UnifiedVectorFormat rowid_data;
-		rowid_vec.ToUnifiedFormat(chunk.size(), rowid_data);
-		auto rowids = UnifiedVectorFormat::GetData<row_t>(rowid_data);
+	TaskExecutor task_executor(context);
+	for (idx_t i = 0; i < num_tasks; ++i) {
```
dentiny, line 242: How are tasks distributed? Apart from the `local_entries` it seems all other arguments are shared across tasks?

benchmark/benchmark_cache_build.cpp
```diff
@@ -0,0 +1,106 @@
...
+	con.Query(sql); // warmup
```
dentiny: curious warmup for what?

benchmark/benchmark_cache_build.cpp
dentiny: Document on return value

benchmark/benchmark_cache_build.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+	const int rounds = 5;
+	const std::string build = "SELECT * FROM condition_cache_build('t', 'val = 42')";
+
+	// 0. Query baseline vs cache build cost
+	con.Query("CREATE OR REPLACE TABLE t AS SELECT i AS id, i%100 AS val FROM range(2000000) t(i)");
+	std::vector<std::pair<std::string, double>> results;
+	results.emplace_back("SELECT count(*) WHERE val = 42",
+	                     BenchMs(con, "SELECT count(*) FROM t WHERE val = 42", rounds));
+	results.emplace_back("SELECT count(*) WHERE val < 50",
```
dentiny, line 49: completely duplicate sql

src/query_condition_cache_functions.cpp
```diff
@@ -134,41 +216,38 @@ shared_ptr<ConditionCacheEntry> BuildCacheEntry(ClientContext &context, DuckTabl
...
-			}
-			auto row_id = NumericCast<idx_t>(rowids[rowid_entry]);
-			idx_t row_group_idx = row_id / DEFAULT_ROW_GROUP_SIZE;
-			idx_t vector_idx = (row_id % DEFAULT_ROW_GROUP_SIZE) / STANDARD_VECTOR_SIZE;
-			entry->bitvectors[row_group_idx].SetVector(vector_idx);
+	// Merge local entries into final result
+	auto entry = make_shared_ptr<ConditionCacheEntry>();
+	for (const auto &local : local_entries) {
+		for (const auto &pair : local.bitvectors) {
+			entry->bitvectors[pair.first].MergeFrom(pair.second);
```
dentiny: nit: use structural binding to improve readability

src/query_condition_cache_functions.cpp
```diff
@@ -194,7 +273,12 @@ CacheEntryStats ComputeCacheEntryStats(const ConditionCacheEntry &entry, idx_t t
 		total_vectors += (remaining_rows + STANDARD_VECTOR_SIZE - 1) / STANDARD_VECTOR_SIZE;
 	}
 
-	idx_t qualifying_row_groups = entry.bitvectors.size();
+	idx_t qualifying_row_groups = 0;
+	for (const auto &bitvector_pair : entry.bitvectors) {
```
dentiny: can we make it a util function for `ConditionCacheEntry`?

benchmark/benchmark_cache_build.cpp
```diff
@@ -0,0 +1,106 @@
...
+	                     BenchMs(con, "SELECT count(*) FROM t WHERE val < 50", rounds));
+	results.emplace_back("cache build (val = 42)", BenchMs(con, build, rounds));
+	results.emplace_back("cache build (val < 50)",
+	                     BenchMs(con, "SELECT * FROM condition_cache_build('t', 'val < 50')", rounds));
+	PrintSection("Query baseline vs cache build cost (2M rows, threads=default)", results);
+
+	// 1. Row count scaling
+	results.clear();
+	for (int64_t n : {100000, 500000, 2000000, 12288000}) {
+		con.Query("CREATE OR REPLACE TABLE t AS SELECT i AS id, i%100 AS val FROM range(" + std::to_string(n) + ") t(i)");
```
dentiny: Use `StringUtil::Format` instead of `operator+`

dentiny (commented):
second round, will continue later tonight

src/query_condition_cache_functions.cpp
Andrewtangtang: Tasks share arguments are all read-only. Each task calls NextParallelScan() which internally synchronizes
to hand out the next unscanned row group. The only per-task mutable state is local_entries.

src/query_condition_cache_functions.cpp
Andrewtangtang: This instantiates an empty RowGroupFilter for row groups where no rows match the predicate. Without this, those row groups would be absent from the cache, and downstream filters cannot distinguish "not scanned" from "scanned with no matches".

Andrewtangtang:
@dentiny Could you please take another look when you have time? Thanks.

src/query_condition_cache_functions.cpp
dentiny: Thanks!

CMakeLists.txt
```diff
@@ -29,3 +29,8 @@ install(
   ARCHIVE DESTINATION "${INSTALL_LIB_DIR}")
 
 add_subdirectory(test/unittest)
+
+if(BUILD_BENCHMARK)
```
dentiny: - the cmake options is called `BUILD_BENCHMARKS`: https://github.com/duckdb/duckdb/blob/2b227bfeb219991ea4ae809598ab7be242c7e85d/CMakeLists.txt#L481
- I'm not familiar with cmake syntax, but seems you need to evaluate it with `${BUILD_BENCHMARKS}`: https://github.com/duckdb/duckdb/blob/2b227bfeb219991ea4ae809598ab7be242c7e85d/CMakeLists.txt#L907

Makefile
```diff
@@ -30,4 +30,9 @@ format-all: format
 	cmake-format -i CMakeLists.txt
 	cmake-format -i test/unittest/CMakeLists.txt
 
-PHONY: format-all test_release_all test_debug_all test_reldebug_all
+benchmark_cache_build:
+	cmake -DBUILD_BENCHMARK=ON -B build/release -S duckdb
```
dentiny: oh it is intentional to use a different option as DuckDB (BUILD_BENCHMARK VS BUILD_BENCHMARKS)? if so, why?

dentiny (approved):
Thanks!

PR #43, opened 2026-03-27
Add CacheExpressionFilter for cache-based pruning

Andrewtangtang:
## Summary

This PR adds `CacheExpressionFilter`, a two-level filter mechanism that uses cached bitvectors to skip scanning data that has no qualifying rows.

- **Row-group level** (`CheckStatistics`): If all row groups in the stats range are cached and have no matches, returns `FILTER_ALWAYS_FALSE` to skip the entire row group.
- **Vector level** (`ConditionCacheFilterFn`): For each vector, checks the bitvector and returns a constant boolean — `false` skips the vector entirely.

The filter defaults to pass-through for row groups not in the cache (e.g. newly inserted data) to ensure correctness.

## Note
 This PR only implements the filter logic. It is not yet injected into query plans — the upcoming `QueryConditionCacheOptimizer` PR will inject `CacheExpressionFilter` into `LogicalGet` nodes during the post-optimize phase, making the cache automatically applied when `SET use_query_condition_cache = true`.

### DIFF
--- a/src/include/query_condition_cache_filter.hpp
+++ b/src/include/query_condition_cache_filter.hpp
@@ -1,0 +1,52 @@
+#pragma once
+
+#include "query_condition_cache_state.hpp"
+
+#include "duckdb/function/scalar_function.hpp"
+#include "duckdb/planner/expression.hpp"
+#include "duckdb/planner/filter/expression_filter.hpp"
+#include "duckdb/storage/statistics/numeric_stats.hpp"
+
+namespace duckdb {
+
+// Holds a shared_ptr to the ConditionCacheEntry so the filter function
+// and CheckStatistics can access the bitvectors during execution.
+struct ConditionCacheFilterBindData : public FunctionData {
+	shared_ptr<ConditionCacheEntry> cache_entry;
+
+	explicit ConditionCacheFilterBindData(shared_ptr<ConditionCacheEntry> entry);
+
+	unique_ptr<FunctionData> Copy() const override;
+	bool Equals(const FunctionData &other_p) const override;
+};
+
+// Called during plan deserialization/verification when the optimizer's bind data
+// is not available. Returns an empty cache entry that passes all rows through.
+unique_ptr<FunctionData> ConditionCacheFilterBind(ClientContext &context, ScalarFunction &bound_function,
+                                                  vector<unique_ptr<Expression>> &arguments);
+
+// Per-thread local state (placeholder for future use)
+struct ConditionCacheFilterState : public FunctionLocalState {};
+
+unique_ptr<FunctionLocalState> ConditionCacheFilterInit(ExpressionState &state, const BoundFunctionExpression &expr,
+                                                        FunctionData *bind_data);
+
+// Vector-level filter: takes a ROW_ID column as input, looks up the bitvector
+// for that row group + vector index, and returns a constant BOOLEAN for the
+// entire vector. Defaults to true for row groups not in cache.
+void ConditionCacheFilterFn(DataChunk &args, ExpressionState &state, Vector &result);
+
+// Row-group level pruning via CheckStatistics. If all row groups covered by
+// the stats range are cached and empty, returns FILTER_ALWAYS_FALSE to skip
+// the entire row group. Otherwise returns NO_PRUNING_POSSIBLE.
+class CacheExpressionFilter : public ExpressionFilter {
+public:
+	shared_ptr<ConditionCacheEntry> cache_entry;
+
+	CacheExpressionFilter(unique_ptr<Expression> expr, shared_ptr<ConditionCacheEntry> entry);
+
+	FilterPropagateResult CheckStatistics(BaseStatistics &stats) const override;
+	unique_ptr<TableFilter> Copy() const override;
+};
+
+} // namespace duckdb
--- a/src/include/query_condition_cache_filter.hpp
+++ b/src/include/query_condition_cache_filter.hpp
@@ -25,9 +25,6 @@ struct ConditionCacheFilterBindData : public FunctionData {
 unique_ptr<FunctionData> ConditionCacheFilterBind(ClientContext &context, ScalarFunction &bound_function,
                                                   vector<unique_ptr<Expression>> &arguments);
 
-// Per-thread local state (placeholder for future use)
-struct ConditionCacheFilterState : public FunctionLocalState {};
-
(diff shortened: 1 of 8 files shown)


src/query_condition_cache_filter.cpp
```diff
@@ -0,0 +1,90 @@
...
+    : cache_entry(std::move(entry)) {
+}
+
+unique_ptr<FunctionData> ConditionCacheFilterBindData::Copy() const {
+	return make_uniq<ConditionCacheFilterBindData>(cache_entry);
+}
+
+bool ConditionCacheFilterBindData::Equals(const FunctionData &other_p) const {
+	auto &other = other_p.Cast<ConditionCacheFilterBindData>();
+	return cache_entry == other.cache_entry;
```
dentiny, line 19: just to confirm -- it means two shared pointer pointing to the same object

src/query_condition_cache_filter.cpp
```diff
@@ -0,0 +1,90 @@
...
+	input_vec.ToUnifiedFormat(args.size(), vdata);
+	auto row_ids = UnifiedVectorFormat::GetData<int64_t>(vdata);
+
+	auto first_idx = vdata.sel->get_index(0);
+	int64_t first_row_id = row_ids[first_idx];
+
+	idx_t rg_idx = NumericCast<idx_t>(first_row_id) / DEFAULT_ROW_GROUP_SIZE;
+	idx_t vec_idx = (NumericCast<idx_t>(first_row_id) % DEFAULT_ROW_GROUP_SIZE) / STANDARD_VECTOR_SIZE;
+	bool passes = true;
+	auto it = entry.bitvectors.find(rg_idx);
```
dentiny, line 59: Could the bitvectors be updated? https://github.com/dentiny/duckdb-query-condition-cache/blob/76007284c7d07165b54470bb3e0841fedc6b7471/src/include/query_condition_cache_state.hpp#L60

If so, we should add mutex; otherwise let's mark it as `const`

src/query_condition_cache_filter.cpp
```diff
@@ -0,0 +1,90 @@
...
+unique_ptr<FunctionLocalState> ConditionCacheFilterInit(ExpressionState &state, const BoundFunctionExpression &expr,
+                                                        FunctionData *bind_data) {
+	return make_uniq<ConditionCacheFilterState>();
+}
+
+void ConditionCacheFilterFn(DataChunk &args, ExpressionState &state, Vector &result) {
+	auto &bind_data = state.expr.Cast<BoundFunctionExpression>().bind_info->Cast<ConditionCacheFilterBindData>();
+	auto &entry = *bind_data.cache_entry;
+
+	auto &input_vec = args.data[0];
```
dentiny, line 45: Let's add an assertion or exception check here, otherwise invalid memory access or UB

src/query_condition_cache_filter.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+	// All row_ids in a single vector belong to the same row group + vector range,
+	// so checking the first row_id is sufficient.
+	UnifiedVectorFormat vdata;
+	input_vec.ToUnifiedFormat(args.size(), vdata);
+	auto row_ids = UnifiedVectorFormat::GetData<int64_t>(vdata);
+
+	auto first_idx = vdata.sel->get_index(0);
```
dentiny, line 53: Let's add an assertion or exception check here, otherwise invalid memory access or UB

src/include/query_condition_cache_filter.hpp
```diff
@@ -0,0 +1,52 @@
...
+// Per-thread local state (placeholder for future use)
+struct ConditionCacheFilterState : public FunctionLocalState {};
+
+unique_ptr<FunctionLocalState> ConditionCacheFilterInit(ExpressionState &state, const BoundFunctionExpression &expr,
+                                                        FunctionData *bind_data);
+
+// Vector-level filter: takes a ROW_ID column as input, looks up the bitvector
+// for that row group + vector index, and returns a constant BOOLEAN for the
+// entire vector. Defaults to true for row groups not in cache.
+void ConditionCacheFilterFn(DataChunk &args, ExpressionState &state, Vector &result);
```
dentiny, line 34: I guess it will be hooked into optimizer, right?

src/query_condition_cache_filter.cpp
Andrewtangtang: Yes, pointer comparison — returns true only when both point to the same `ConditionCacheEntry` object.

src/include/query_condition_cache_filter.hpp
Andrewtangtang: Yes, it will be injected by QueryConditionCacheOptimizer in the post-optimize phase.

src/query_condition_cache_filter.cpp
Andrewtangtang: I add a mutex, because it may be erased by the invalidation.

dentiny: https://github.com/dentiny/duckdb-query-condition-cache/pull/47 I'm wondering if we could use annotated mutex, so compilation fails/warns if we misuse (i.e., forget to acquire lock)?

Andrewtangtang: I added it.

src/include/query_condition_cache_filter.hpp
```diff
@@ -0,0 +1,52 @@
...
+#include "duckdb/function/scalar_function.hpp"
+#include "duckdb/planner/expression.hpp"
+#include "duckdb/planner/filter/expression_filter.hpp"
+#include "duckdb/storage/statistics/numeric_stats.hpp"
+
+namespace duckdb {
+
+// Holds a shared_ptr to the ConditionCacheEntry so the filter function
+// and CheckStatistics can access the bitvectors during execution.
+struct ConditionCacheFilterBindData : public FunctionData {
```
dentiny, line 14: I'm wondering if we need to put ConditionCacheFilterBindData / ConditionCacheFilterState in the hpp file?

dentiny (approved):
Feel free to merge after addressing the comment. Thanks!

src/include/query_condition_cache_filter.hpp
Andrewtangtang: Moved ConditionCacheFilterState to `.cpp`. ConditionCacheFilterBindData stays in the header since the optimizer's InjectCacheFilter needs it.

PR #44, opened 2026-03-28
Add optimizer pre-optimize and predicate key normalization

Andrewtangtang:
## Summary

This PR adds the optimizer pre-optimize phase for automatic predicate caching. When `use_query_condition_cache` is enabled, queries with `WHERE` clauses automatically build cache entries without manual `condition_cache_build` calls.

- Add `QueryConditionCacheOptimizer` that runs before FilterPushdown to capture the full WHERE clause, detect `LogicalFilter -> LogicalGet` patterns, and build cache entries inline on cache miss
- Add `NormalizeExpressionForKey` to canonicalize comparison operand order (constants on the right) so equivalent predicates share the same cache key
- Rename setting from `enable_query_condition_cache` to `use_query_condition_cache` 

This PR only builds the cache — it does not yet apply it during scans. A follow-up PR will add the post-optimize phase to inject `CacheExpressionFilter` for row-group/vector skipping.

### DIFF
--- a/src/include/predicate_key_utils.hpp
+++ b/src/include/predicate_key_utils.hpp
@@ -7,12 +7,14 @@ namespace duckdb {
 class DuckTableEntry;
 
 // Parse a predicate SQL string, bind it against a table using CheckBinder,
-// and return the bound expression. Returns nullptr if parsing fails.
+// and return the bound expression. Returns nullptr if predicate_sql is empty
+// or parsing fails.
 unique_ptr<Expression> BindPredicate(ClientContext &context, DuckTableEntry &table_entry, const string &predicate_sql);
 
 // Compute a canonical cache key string from a predicate SQL.
 // Parses, binds, normalizes (comparison operand order + conjunction sort),
-// and returns the normalized ToString(). Returns empty string on failure.
+// and returns the normalized ToString(). Returns empty string if predicate_sql
+// is empty or parsing fails.
 string ComputeCanonicalPredicateKey(ClientContext &context, DuckTableEntry &table_entry, const string &predicate_sql);
 
 // Normalize a bound expression tree in-place for canonical cache key generation.
--- a/src/include/predicate_key_utils.hpp
+++ b/src/include/predicate_key_utils.hpp
@@ -1,0 +1,24 @@
+#pragma once
+
+#include "duckdb/planner/expression.hpp"
+
+namespace duckdb {
+
+class DuckTableEntry;
+
+// Parse a predicate SQL string, bind it against a table using CheckBinder,
+// and return the bound expression. Returns nullptr if parsing fails.
+unique_ptr<Expression> BindPredicate(ClientContext &context, DuckTableEntry &table_entry, const string &predicate_sql);
+
+// Compute a canonical cache key string from a predicate SQL.
+// Parses, binds, normalizes (comparison operand order + conjunction sort),
+// and returns the normalized ToString(). Returns empty string on failure.
+string ComputeCanonicalPredicateKey(ClientContext &context, DuckTableEntry &table_entry, const string &predicate_sql);
+
+// Normalize a bound expression tree in-place for canonical cache key generation.
+// Ensures equivalent predicates produce the same ToString() output:
+//   - Comparison operands: constants moved to the right ("42 = val" -> "val = 42")
+//   - Conjunction children (AND/OR): sorted by ToString() ("b = 2 OR a = 1" -> "a = 1 OR b = 2")
+void NormalizeExpressionForCacheKey(Expression &expr);
+
+} // namespace duckdb
--- a/src/include/query_condition_cache_functions.hpp
+++ b/src/include/query_condition_cache_functions.hpp
@@ -9,12 +9,6 @@ namespace duckdb {
 class DuckTableEntry;
 class Expression;
 
-// Normalize a bound expression tree in-place for canonical cache key generation.
-// Ensures equivalent predicates produce the same ToString() output:
-//   - Comparison operands: constants moved to the right ("42 = val" → "val = 42")
-//   - Conjunction children (AND/OR): sorted by ToString() ("b = 2 OR a = 1" -> "a = 1 OR b = 2")
-void NormalizeExpressionForKey(Expression &expr);
-
 // TODO: Exposed for future reuse and C++ unit testing.
 // Scan all rows in the table, evaluate bound_expr, and build a ConditionCacheEntry
 // recording which vectors contain qualifying rows.
--- a/src/include/query_condition_cache_functions.hpp
+++ b/src/include/query_condition_cache_functions.hpp
@@ -9,9 +9,10 @@ namespace duckdb {
 class DuckTableEntry;
 class Expression;
 
-// Normalize comparison operand order in-place: put foldable constants on the right
-// so that "42 = val" and "val = 42" produce the same ToString() for cache key.
-// Reference: duckdb/src/optimizer/rule/comparison_simplification.cpp
+// Normalize a bound expression tree in-place for canonical cache key generation.
+// Ensures equivalent predicates produce the same ToString() output:
+//   - Comparison operands: constants moved to the right ("42 = val" → "val = 42")
+//   - Conjunction children (AND/OR): sorted by ToString() ("b = 2 OR a = 1" -> "a = 1 OR b = 2")
 void NormalizeExpressionForKey(Expression &expr);
 
 // TODO: Exposed for future reuse and C++ unit testing.
--- a/src/include/query_condition_cache_functions.hpp
+++ b/src/include/query_condition_cache_functions.hpp
@@ -9,6 +9,11 @@ namespace duckdb {
 class DuckTableEntry;
 class Expression;
 
+// Normalize comparison operand order in-place: put foldable constants on the right
+// so that "42 = val" and "val = 42" produce the same ToString() for cache key.
+// Reference: duckdb/src/optimizer/rule/comparison_simplification.cpp
+void NormalizeExpressionForKey(Expression &expr);
+
 // TODO: Exposed for future reuse and C++ unit testing.
 // Scan all rows in the table, evaluate bound_expr, and build a ConditionCacheEntry
 // recording which vectors contain qualifying rows.
--- a/src/include/query_condition_cache_optimizer.hpp
+++ b/src/include/query_condition_cache_optimizer.hpp
@@ -1,0 +1,52 @@
+#pragma once
+
+#include "query_condition_cache_state.hpp"
+
+#include "duckdb/main/client_context_state.hpp"
+#include "duckdb/optimizer/optimizer_extension.hpp"
+
+namespace duckdb {
+
+class LogicalGet;
+
+// Query-scoped state for passing cache entries between pre-optimize and post-optimize phases.
+// Stored in ClientContext::registered_state; automatically cleared on QueryEnd.
+struct CacheOptimizerQueryState : public ClientContextState {
+	// Maps table_index -> cache entry for tables matched during pre-optimize.
+	// Consumed by post-optimize to inject cache filters.
+	unordered_map<idx_t, shared_ptr<ConditionCacheEntry>> cache_apply_pending;
+
+	void QueryEnd(ClientContext &context, optional_ptr<ErrorData> error) override {
+		cache_apply_pending.clear();
+	}
+};
+
+class QueryConditionCacheOptimizer : public OptimizerExtension {
+public:
+	QueryConditionCacheOptimizer();
+
+	// Pre-optimize: compute canonical predicate keys before FilterPushdown splits the WHERE clause.
+	// On cache miss, builds cache inline so the first query benefits immediately.
+	static void PreOptimizeFunction(OptimizerExtensionInput &input, unique_ptr<LogicalOperator> &plan);
+
+	// Post-optimize: inject cache filters into LogicalGet nodes that were matched pre-optimize.
+	static void OptimizeFunction(OptimizerExtensionInput &input, unique_ptr<LogicalOperator> &plan);
+
+private:
+	static bool IsSettingEnabled(ClientContext &context);
+
+	// Walk plan before FilterPushdown: find LogicalFilter -> LogicalGet, compute key, lookup/build cache
+	static void PreOptimizeWalk(ClientContext &context, unique_ptr<LogicalOperator> &plan, bool inside_dml,
+	                            CacheOptimizerQueryState &state);
+
+	// Compute a canonical predicate key via parse+bind normalization,
+	// matching the key format used by condition_cache_build/info.
+	static CacheKey ComputePredicateKey(ClientContext &context, idx_t table_oid,
+	                                    const vector<unique_ptr<Expression>> &expressions, LogicalGet &get);
+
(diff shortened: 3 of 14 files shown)


src/include/query_condition_cache_optimizer.hpp
```diff
@@ -0,0 +1,52 @@
...
+#include "duckdb/optimizer/optimizer_extension.hpp"
+
+namespace duckdb {
+
+class LogicalGet;
+
+// Query-scoped state for passing cache entries between pre-optimize and post-optimize phases.
+// Stored in ClientContext::registered_state; automatically cleared on QueryEnd.
+struct CacheOptimizerQueryState : public ClientContextState {
+	// Maps table_index -> cache entry for tables matched during pre-optimize.
```
dentiny, line 18: may I ask what's table index?

src/query_condition_cache_extension.cpp
```diff
@@ -2,31 +2,29 @@
...
 void LoadInternal(ExtensionLoader &loader) {
 	loader.RegisterFunction(ConditionCacheBuildFunction());
 	loader.RegisterFunction(ConditionCacheInfoFunction());
 
-	// Register extension settings
-	auto &db_instance = loader.GetDatabaseInstance();
-	auto &config = DBConfig::GetConfig(db_instance);
-	config.AddExtensionOption("enable_query_condition_cache", "Enable or disable the query condition cache feature",
-	                          LogicalType {LogicalTypeId::BOOLEAN}, Value(true), EnableQueryConditionCacheCallback);
+	// Register the use_query_condition_cache setting (default: false)
```
dentiny: Hmm why false? I think the extension should enable by default, and provide option to disable.

src/include/query_condition_cache_functions.hpp
```diff
@@ -9,6 +9,12 @@ namespace duckdb {
 class DuckTableEntry;
 class Expression;
 
+// Normalize a bound expression tree in-place for canonical cache key generation.
+// Ensures equivalent predicates produce the same ToString() output:
+//   - Comparison operands: constants moved to the right ("42 = val" → "val = 42")
+//   - Conjunction children (AND/OR): sorted by ToString() ("b = 2 OR a = 1" -> "a = 1 OR b = 2")
+void NormalizeExpressionForKey(Expression &expr);
```
dentiny:
```suggestion
void NormalizeExpressionForCacheKey(Expression &expr);
```

test/unittest/test_normalize_expression.cpp
```diff
@@ -0,0 +1,128 @@
...
+	physical_index_set_t bound_columns;
+	CheckBinder check_binder(*binder, context, table_entry.name, table_entry.GetColumns(), bound_columns);
+	auto bound_expr = check_binder.Bind(parsed_exprs[0]);
+	NormalizeExpressionForKey(*bound_expr);
+	return bound_expr->ToString();
+}
+
+} // namespace
+
+TEST_CASE("NormalizeExpressionForKey - comparison operand order", "[normalize]") {
```
dentiny: A few missing unit test cases:
- with legal space in the statement: val = 42 & val =    42
- both sides foldable: 1 = 2 & 2 = 1
- more than two conjunction children: a=1 AND b=2 AND c=3 & c=3 AND a=1 AND b=2

src/query_condition_cache_functions.cpp
```diff
@@ -339,6 +370,7 @@ unique_ptr<FunctionData> ConditionCacheInfoBind(ClientContext &context, TableFun
 		physical_index_set_t bound_columns;
 		CheckBinder check_binder(*binder, context, qname.name, table_entry.GetColumns(), bound_columns);
 		auto bound_expr = check_binder.Bind(parsed_exprs[0]);
+		NormalizeExpressionForKey(*bound_expr);
 		result->canonical_key = bound_expr->ToString();
```
dentiny: I'm wondering if we could 
- move NormalizeExpressionForKey into a separate util file
- it's easy to forget canonicalization since we have to manually call it every single time, any chance we could provide a util function, which takes a key and we always to canonicalization inside?

src/query_condition_cache_optimizer.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+// Pre-optimize runs BEFORE built-in passes (including FilterPushdown).
+// The full WHERE clause is still in LogicalFilter.expressions, so we can
+// compute a canonical key that covers the entire predicate.
+void QueryConditionCacheOptimizer::PreOptimizeFunction(OptimizerExtensionInput &input,
+                                                       unique_ptr<LogicalOperator> &plan) {
+	if (!IsSettingEnabled(input.context)) {
+		return;
+	}
+	auto query_state = input.context.registered_state->GetOrCreate<CacheOptimizerQueryState>("qcc_optimizer_state");
```
dentiny: nit: better make it a const `qcc_optimizer_state`, not sure if it'd be used somewhere else

src/query_condition_cache_optimizer.cpp
```diff
@@ -0,0 +1,190 @@
...
+		return; // not a DuckDB table (e.g. system table, external)
+	}
+	if (filter.expressions.empty()) {
+		return;
+	}
+
+	auto &duck_table = table->Cast<DuckTableEntry>();
+	auto &storage = duck_table.GetStorage();
+
+	// Skip caching for tables with indexes — index scans are typically more efficient
```
dentiny: I'm not sure about this, do you mind leaving a TODO for later revisit?
For example, with condition cache we could potentially skip an entire table

src/query_condition_cache_optimizer.cpp
```diff
@@ -0,0 +1,190 @@
...
+	if (key.filter_key.empty()) {
+		return;
+	}
+
+	auto store = ConditionCacheStore::GetOrCreate(context);
+	auto entry = store->Lookup(context, key);
+
+	if (!entry) {
+		// Cache miss: build cache inline
+		entry = BuildCacheForPredicate(context, filter.expressions, get);
```
dentiny, line 92: Please leave a TODO: this function runs long since we scan all rows and build cache, I'm thinking if we could build cache in the background and sync to cache later.

src/query_condition_cache_optimizer.cpp
```diff
@@ -0,0 +1,190 @@
...
+			store->Upsert(context, key, entry);
+		}
+	}
+
+	if (entry) {
+		state.cache_apply_pending[get.table_index] = std::move(entry);
+	}
+}
+
+CacheKey QueryConditionCacheOptimizer::ComputePredicateKey(ClientContext &context, idx_t table_oid,
```
dentiny: Let's leave a TODO: DuckDB core upgrades to C++17 in main, which will be released in v2.0, let's use std::optional then. Checking whether filter key is empty is not good.

src/query_condition_cache_optimizer.cpp
```diff
@@ -0,0 +1,190 @@
...
+                                                           const vector<unique_ptr<Expression>> &expressions,
+                                                           LogicalGet &get) {
+	auto table_ptr = get.GetTable();
+	if (!table_ptr) {
+		return CacheKey {table_oid, ""};
+	}
+	auto &table_entry = table_ptr->Cast<DuckTableEntry>();
+
+	// Reconstruct SQL from expressions, sort for canonical ordering
```
dentiny:
```suggestion
	// Reconstruct SQL from expressions, sort for alphabetical ordering
```
?

src/query_condition_cache_optimizer.cpp
```diff
@@ -0,0 +1,190 @@
...
+	vector<string> parts;
+	parts.reserve(expressions.size());
```
dentiny, line 106: https://github.com/dentiny/duckdb-query-condition-cache/blob/9439ac1256aca67c2c212e21bae01663dc249d39/src/query_condition_cache_optimizer.cpp#L126-L133
We don't seem to sort here ^, is it intentional?

dentiny: Can we extract it out into a util function (and put into a util file)?

dentiny: Actually ComputePredicateKey and BuildCacheForPredicate seems overlap a lot: ToString → Parse → Bind -> xxx, can we share common utils?

src/query_condition_cache_functions.cpp
```diff
@@ -26,6 +28,34 @@
...
+		}
+		return;
+	}
+
+	// Sort conjunction children for canonical ordering
+	if (expr.GetExpressionClass() == ExpressionClass::BOUND_CONJUNCTION) {
+		auto &conj = expr.Cast<BoundConjunctionExpression>();
+		sort(conj.children.begin(), conj.children.end(),
+		     [](const unique_ptr<Expression> &a, const unique_ptr<Expression> &b) {
+			     return a->ToString() < b->ToString();
```
dentiny: leave a TODO please, `ToString` is called O(NlogN) times when sorting

src/query_condition_cache_optimizer.cpp
```diff
@@ -0,0 +1,190 @@
...
+	}
+	auto query_state = input.context.registered_state->GetOrCreate<CacheOptimizerQueryState>("qcc_optimizer_state");
+	query_state->cache_apply_pending.clear();
+	try {
+		PreOptimizeWalk(input.context, plan, /*inside_dml=*/false, *query_state);
+	} catch (...) {
+		// If anything goes wrong, clear pending entries and let the query
+		// proceed without cache optimization
+		query_state->cache_apply_pending.clear();
+	}
```
peterxcli: let's remove the exception ignore for now, if there are really some query we cant process, we should then add the gate rule code explictly.
```suggestion
	PreOptimizeWalk(input.context, plan, /*inside_dml=*/false, *query_state);
```

src/query_condition_cache_optimizer.cpp
peterxcli:
```suggestion
	auto query_state = input.context.registered_state->GetOrCreate<CacheOptimizerQueryState>(CacheOptimizerQueryState::Name);
```
can we do this?

src/query_condition_cache_optimizer.cpp
```diff
@@ -0,0 +1,190 @@
...
+		return;
+	}
+
+	auto &duck_table = table->Cast<DuckTableEntry>();
+	auto &storage = duck_table.GetStorage();
+
+	// Skip caching for tables with indexes — index scans are typically more efficient
+	if (storage.HasIndexes()) {
+		return;
+	}
```
peterxcli: sorry let's remove this for now, I'm not sure if this claim is right.

peterxcli (commented):
others look good

src/include/query_condition_cache_optimizer.hpp
Andrewtangtang: `table_index` is `LogicalGet.table_index` a per-query temporary ID assigned during planning. I use it to match LogicalGet nodes between pre-optimize and post-optimize phases. Using `table_oid` wouldn't work for self-joins where the same table appears multiple times.

dentiny: Thanks!

src/include/predicate_key_utils.hpp
```diff
@@ -0,0 +1,24 @@
...
+
+#include "duckdb/planner/expression.hpp"
+
+namespace duckdb {
+
+class DuckTableEntry;
+
+// Parse a predicate SQL string, bind it against a table using CheckBinder,
+// and return the bound expression. Returns nullptr if parsing fails.
+unique_ptr<Expression> BindPredicate(ClientContext &context, DuckTableEntry &table_entry, const string &predicate_sql);
```
dentiny, line 12: curious what if the predicate sql is empty string?
(I mean test + comment update)

src/predicate_key_utils.cpp
```diff
@@ -0,0 +1,69 @@
...
+// (duckdb/src/optimizer/rule/comparison_simplification.cpp) since
+// ExpressionRewriter runs after our pre-optimize pass.
+// Bottom-up: normalize children before sorting conjunction children.
+void NormalizeExpressionForCacheKey(Expression &expr) {
+	ExpressionIterator::EnumerateChildren(expr, [](Expression &child) { NormalizeExpressionForCacheKey(child); });
+
+	// Normalize comparison operand order
+	if (expr.GetExpressionClass() == ExpressionClass::BOUND_COMPARISON) {
+		auto &comp = expr.Cast<BoundComparisonExpression>();
+		if (comp.left->IsFoldable() && !comp.right->IsFoldable()) {
```
dentiny: what if both non-foldable?

src/predicate_key_utils.cpp
```diff
@@ -0,0 +1,69 @@
...
+			if (comp.left->ToString() > comp.right->ToString()) {
+				std::swap(comp.left, comp.right);
+				comp.type = FlipComparisonExpression(comp.type);
+			}
+		}
+		return;
+	}
+
+	// Sort conjunction children for canonical ordering
+	// TODO: ToString() is called O(NlogN) times during sort
```
dentiny, line 58: `ToString` is used very frequently in our predicate cache entry construction, I somehow feel like we need to have a small LRU cache

src/query_condition_cache_extension.cpp
dentiny: @Andrewtangtang Any explanation? Is it for dev stage only?

src/query_condition_cache_functions.cpp
```diff
@@ -268,28 +269,20 @@ void ConditionCacheBuildExecute(ClientContext &context, TableFunctionInput &data
 	auto &table_entry =
 	    Catalog::GetEntry<DuckTableEntry>(context, bind_data.catalog, bind_data.schema, bind_data.table);
 
-	// Parse predicate SQL and bind column references to table columns
-	auto parsed_exprs = Parser::ParseExpressionList(bind_data.predicate_sql);
-	if (parsed_exprs.empty()) {
+	string canonical_key = ComputeCanonicalPredicateKey(context, table_entry, bind_data.predicate_sql);
```
dentiny, line 272: Let's add a corner case what predicate is empty string, or empty string after stripping.

src/query_condition_cache_extension.cpp
Andrewtangtang: sorry. I forgot to update comment.

src/include/predicate_key_utils.hpp
Andrewtangtang: Added guard in BindPredicate — returns nullptr for empty or whitespace-only predicate SQL.

src/predicate_key_utils.cpp
Andrewtangtang: Both non-foldable operands (e.g. id = val vs val = id) are now normalized by sorting via ToString(). Same logic applies to both-foldable. Added test.

src/query_condition_cache_functions.cpp
Andrewtangtang: Handled — BindPredicate guards against empty/whitespace-only strings. condition_cache_build throws "failed to parse predicate".

src/predicate_key_utils.cpp
Andrewtangtang: Agreed, will address in a follow-up PR.

dentiny (approved):
Thanks!

PR #45, opened 2026-03-29
Add cache invalidation support to store

Andrewtangtang:
## Summary

This PR adds cache invalidation support to `ConditionCacheStore`, enabling row-group-level removal of cached predicate bitvectors when table data changes.

- `TableFilterKeyIndex`: A per-table reverse index (stored in `ObjectCache`) that tracks which `filter_key`s have cache entries for a given table OID. This enables efficient lookup of all cache entries affected by a table mutation without scanning the entire cache.
- `RemoveRowGroupsForTable(context, table_oid, row_group_indices)`: Removes the specified row groups from all cache entries belonging to a table. If an entry has no remaining row groups after removal, the entry and its index reference are cleaned up entirely.
- `HasEntriesForTable(context, table_oid)`: Quick check for whether any cache entries exist for a table.

When DML operations (INSERT, DELETE, UPDATE, VACUUM) modify a table, the affected row groups' cached bitvectors become stale. This PR provides the invalidation primitives that future DML hooks will use to call `RemoveRowGroupsForTable` with the touched row group indices.

## Related Issues

- None

### DIFF
--- a/src/include/query_condition_cache_state.hpp
+++ b/src/include/query_condition_cache_state.hpp
@@ -84,7 +84,7 @@ struct ConditionCacheEntry : public ObjectCacheEntry {
 // invalidation lookup when DML modifies the table.
 struct TableFilterKeyIndex : public ObjectCacheEntry {
 	concurrency::mutex lock;
-	vector<string> filter_keys DUCKDB_GUARDED_BY(lock);
+	unordered_set<string> filter_keys DUCKDB_GUARDED_BY(lock);
 
 	static string ObjectType() {
 		return "query_condition_cache_filter_key_index";
@@ -98,10 +98,12 @@ struct TableFilterKeyIndex : public ObjectCacheEntry {
 		return optional_idx {};
 	}
 
+	// Add a filter key. No-op if it already exists (deduplication).
 	void Add(const string &filter_key);
+	// Remove a filter key. No-op if it doesn't exist.
 	void Remove(const string &filter_key);
 	bool IsEmpty();
-	vector<string> GetAll();
+	unordered_set<string> GetAll();
 };
 
 // Stored in DuckDB's per-database ObjectCache
--- a/src/include/query_condition_cache_state.hpp
+++ b/src/include/query_condition_cache_state.hpp
@@ -83,8 +83,8 @@ struct ConditionCacheEntry : public ObjectCacheEntry {
 // filter_keys have cache entries for a given table, enabling fast
 // invalidation lookup when DML modifies the table.
 struct TableFilterKeyIndex : public ObjectCacheEntry {
-	mutex lock;
-	vector<string> filter_keys;
+	concurrency::mutex lock;
+	vector<string> filter_keys DUCKDB_GUARDED_BY(lock);
 
 	static string ObjectType() {
 		return "query_condition_cache_filter_key_index";
@@ -133,13 +133,18 @@ public:
 	// Check if any entries exist for a given table OID
 	bool HasEntriesForTable(ClientContext &context, idx_t table_oid);
 
+	// Clear all cache entries and filter key indexes
+	void ClearAll(ClientContext &context);
+
 	// Get or create the store from a client context
 	static shared_ptr<ConditionCacheStore> GetOrCreate(ClientContext &context);
 
 private:
-	// Generate a unique cache key string from CacheKey
+	concurrency::mutex lock;
+	// Tracks all table OIDs that have been cached, for ClearAll
+	unordered_set<idx_t> cached_table_oids DUCKDB_GUARDED_BY(lock);
+
 	static string MakeCacheKeyString(const CacheKey &key);
-	// Generate a unique filter key index key from table OID
 	static string MakeFilterKeyIndexKey(idx_t table_oid);
 };
 
--- a/src/include/query_condition_cache_state.hpp
+++ b/src/include/query_condition_cache_state.hpp
@@ -84,6 +84,8 @@ struct ConditionCacheEntry : public ObjectCacheEntry {
 // invalidation lookup when DML modifies the table.
 struct TableFilterKeyIndex : public ObjectCacheEntry {
 	concurrency::mutex lock;
+	// Keys that have been cached for this table; entries may have been evicted by LRU.
+	// Absence of a key guarantees it is not cached.
 	unordered_set<string> filter_keys DUCKDB_GUARDED_BY(lock);
 
 	static string ObjectType() {
@@ -98,9 +100,9 @@ struct TableFilterKeyIndex : public ObjectCacheEntry {
 		return optional_idx {};
 	}
 
-	// Add a filter key. No-op if it already exists (deduplication).
+	// Add a filter key. No-op if it already exists.
 	void Add(const string &filter_key);
-	// Remove a filter key. No-op if it doesn't exist.
+	// Remove a filter key. Assumes the key must appear in the set.
 	void Remove(const string &filter_key);
 	bool IsEmpty();
 	unordered_set<string> GetAll();
--- a/src/include/query_condition_cache_state.hpp
+++ b/src/include/query_condition_cache_state.hpp
@@ -100,6 +100,7 @@ struct TableFilterKeyIndex : public ObjectCacheEntry {
 
 	void Add(const string &filter_key);
 	void Remove(const string &filter_key);
+	bool IsEmpty();
 	vector<string> GetAll();
 };
 
@@ -133,7 +134,7 @@ public:
 	// Check if any entries exist for a given table OID
 	bool HasEntriesForTable(ClientContext &context, idx_t table_oid);
 
-	// Clear all cache entries and filter key indexes
+	// Clear all cache entries and filter key indices
 	void ClearAll(ClientContext &context);
 
 	// Get or create the store from a client context
--- a/src/include/query_condition_cache_state.hpp
+++ b/src/include/query_condition_cache_state.hpp
@@ -4,6 +4,7 @@
 #include "duckdb/common/mutex.hpp"
 #include "duckdb/common/types/hash.hpp"
 #include "duckdb/common/unordered_map.hpp"
+#include "duckdb/common/unordered_set.hpp"
 #include "duckdb/storage/object_cache.hpp"
 
 namespace duckdb {
@@ -74,6 +75,30 @@ struct ConditionCacheEntry : public ObjectCacheEntry {
 	CacheEntryStats ComputeStats(idx_t total_rows) const;
 };
 
+// Per-table index stored in ObjectCache (non-evictable). Tracks which
+// filter_keys have cache entries for a given table, enabling fast
+// invalidation lookup when DML modifies the table.
+struct TableFilterKeyIndex : public ObjectCacheEntry {
+	mutex lock;
+	vector<string> filter_keys;
+
+	static string ObjectType() {
+		return "query_condition_cache_filter_key_index";
+	}
+
+	string GetObjectType() override {
+		return ObjectType();
+	}
+
+	optional_idx GetEstimatedCacheMemory() const override {
+		return optional_idx {};
+	}
+
+	void Add(const string &filter_key);
+	void Remove(const string &filter_key);
+	vector<string> GetAll();
+};
+
 // Stored in DuckDB's per-database ObjectCache
 class ConditionCacheStore : public ObjectCacheEntry {
 public:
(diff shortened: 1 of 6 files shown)


dentiny:
One comment: we need to clear cache when query condition cache setting disabled.

dentiny:
ping me when ready thanks!

dentiny (commented):
A general comment, shall we use annotated mutex whenever possible?

src/include/query_condition_cache_state.hpp
```diff
@@ -101,12 +126,26 @@ class ConditionCacheStore : public ObjectCacheEntry {
...
 	void Upsert(ClientContext &context, const CacheKey &key, shared_ptr<ConditionCacheEntry> entry);
 
+	// Remove specific row groups from all entries for a table. Returns count of row groups removed.
+	idx_t RemoveRowGroupsForTable(ClientContext &context, idx_t table_oid,
+	                              const unordered_set<idx_t> &row_group_indices);
+
+	// Check if any entries exist for a given table OID
+	bool HasEntriesForTable(ClientContext &context, idx_t table_oid);
+
+	// Clear all cache entries and filter key indexes
```
dentiny: nit: can we unify the wording `indices` / `indexes`?

src/query_condition_cache_state.cpp
```diff
@@ -79,26 +79,120 @@ CacheEntryStats ConditionCacheEntry::ComputeStats(idx_t total_rows) const {
...
+		}
+	}
+	filter_keys.push_back(filter_key);
+}
+
+void TableFilterKeyIndex::Remove(const string &filter_key) {
+	lock_guard<mutex> guard(lock);
+	for (auto it = filter_keys.begin(); it != filter_keys.end(); ++it) {
+		if (*it == filter_key) {
+			filter_keys.erase(it);
```
dentiny: It's very inefficient for vector: if we don't care about ordering, we should swap with the last element.

src/query_condition_cache_state.cpp
```diff
@@ -79,26 +79,120 @@ CacheEntryStats ConditionCacheEntry::ComputeStats(idx_t total_rows) const {
...
+	}
+
+	auto filter_keys = index->GetAll();
+	idx_t removed_count = 0;
+
+	for (const auto &filter_key : filter_keys) {
+		CacheKey key {table_oid, filter_key};
+		string cache_key = MakeCacheKeyString(key);
+		auto entry = cache.Get<ConditionCacheEntry>(cache_key);
+		if (!entry) {
```
dentiny, line 153: in which case will the entry be nullptr? please add comment

src/query_condition_cache_state.cpp
```diff
@@ -79,26 +79,120 @@ CacheEntryStats ConditionCacheEntry::ComputeStats(idx_t total_rows) const {
...
+			index->Remove(filter_key);
+		}
+	}
+	return removed_count;
+}
+
+bool ConditionCacheStore::HasEntriesForTable(ClientContext &context, idx_t table_oid) {
+	auto &cache = ObjectCache::GetObjectCache(context);
+	auto index = cache.Get<TableFilterKeyIndex>(MakeFilterKeyIndexKey(table_oid));
+	return index && !index->GetAll().empty();
```
dentiny: `GetAll` is very expensive, you don't want to call it here

src/include/query_condition_cache_state.hpp
```diff
@@ -101,12 +126,26 @@ class ConditionCacheStore : public ObjectCacheEntry {
 	// Upsert an entry
 	void Upsert(ClientContext &context, const CacheKey &key, shared_ptr<ConditionCacheEntry> entry);
 
+	// Remove specific row groups from all entries for a table. Returns count of row groups removed.
+	idx_t RemoveRowGroupsForTable(ClientContext &context, idx_t table_oid,
+	                              const unordered_set<idx_t> &row_group_indices);
+
+	// Check if any entries exist for a given table OID
+	bool HasEntriesForTable(ClientContext &context, idx_t table_oid);
```
dentiny, line 139: Curious when do we plan to call it?

src/query_condition_cache_state.cpp
```diff
@@ -79,26 +79,120 @@ CacheEntryStats ConditionCacheEntry::ComputeStats(idx_t total_rows) const {
...
+}
+
+void ConditionCacheStore::ClearAll(ClientContext &context) {
+	auto &cache = ObjectCache::GetObjectCache(context);
+
+	lock_guard<mutex> guard(lock);
+	for (auto table_oid : cached_table_oids) {
```
dentiny, line 188: do we cleanup `cached_table_oids` outside of the `ClearAll`? What I care about is whether it could go unbounded

dentiny (commented):
We need to followup performance improvement tickets later

test/unittest/test_cache_invalidation.cpp
```diff
@@ -0,0 +1,93 @@
...
+		entry->bitvectors[0].SetVector(0);
+		entry->bitvectors[1].SetVector(0);
+		entry->bitvectors[2].SetVector(0);
+		store->Upsert(context, {1, "val > 5"}, entry);
+
+		unordered_set<idx_t> rg_indices = {0, 2};
+		auto removed = store->RemoveRowGroupsForTable(context, 1, rg_indices);
+		REQUIRE(removed == 2);
+
+		auto found = store->Lookup(context, {1, "val > 5"});
```
dentiny: Any chance we could add comments beside constants?
```cpp
/*arg1=*/1
```

dentiny:
Ping me when ready, thanks!

src/include/query_condition_cache_state.hpp
Andrewtangtang: Use indices.

src/include/query_condition_cache_state.hpp
Andrewtangtang: Called by CacheInvalidationOptimizer when walking DML plans to skip injecting invalidation operators if the table has no cache entries.

src/query_condition_cache_state.cpp
Andrewtangtang: Maybe I could add a check in RemoveRowGroupsForTable. If a table's entries are all removed, we could remove it from `cached_table_oids` as well.

Andrewtangtang:
Feel free to open tickets for improvements. cc @peterxcli

src/include/query_condition_cache_state.hpp
```diff
@@ -78,6 +79,31 @@ struct ConditionCacheEntry : public ObjectCacheEntry {
...
+
+	string GetObjectType() override {
+		return ObjectType();
+	}
+
+	optional_idx GetEstimatedCacheMemory() const override {
+		return optional_idx {};
+	}
+
+	void Add(const string &filter_key);
```
dentiny, line 104: Could you please add documentation what happened if the filter key already exists when add, and doesn't exist when remove?

src/query_condition_cache_state.cpp
```diff
@@ -79,26 +79,128 @@ CacheEntryStats ConditionCacheEntry::ComputeStats(idx_t total_rows) const {
...
+	for (const auto &existing : filter_keys) {
+		if (existing == filter_key) {
+			return;
+		}
+	}
+	filter_keys.push_back(filter_key);
+}
+
+void TableFilterKeyIndex::Remove(const string &filter_key) {
+	concurrency::lock_guard<concurrency::mutex> guard(lock);
```
dentiny, line 90: I'm actually curious in which case would index non-existent? Is it a bug?

src/include/query_condition_cache_state.hpp
```diff
@@ -78,6 +79,31 @@ struct ConditionCacheEntry : public ObjectCacheEntry {
 	CacheEntryStats ComputeStats(idx_t total_rows) const;
 };
 
+// Per-table index stored in ObjectCache (non-evictable). Tracks which
+// filter_keys have cache entries for a given table, enabling fast
+// invalidation lookup when DML modifies the table.
+struct TableFilterKeyIndex : public ObjectCacheEntry {
+	concurrency::mutex lock;
+	vector<string> filter_keys DUCKDB_GUARDED_BY(lock);
```
dentiny: I feel little weird: index usually uses a search-able data structure like tree or hashmap, but here we're using vector.

src/include/query_condition_cache_state.hpp
dentiny: What if we have a bunch of duplicate `filter_key`?

src/query_condition_cache_state.cpp
```diff
@@ -79,26 +79,128 @@ CacheEntryStats ConditionCacheEntry::ComputeStats(idx_t total_rows) const {
...
+		CacheKey key {table_oid, filter_key};
+		string cache_key = MakeCacheKeyString(key);
+		auto entry = cache.Get<ConditionCacheEntry>(cache_key);
+		if (!entry) {
+			// Entry was evicted by LRU, clean up stale index
+			index->Remove(filter_key);
+			continue;
+		}
+		for (auto rg_idx : row_group_indices) {
+			if (entry->bitvectors.erase(rg_idx) > 0) {
```
dentiny:
```suggestion
removed_count += entry->bitvectors.erase(rg_idx);
```

src/query_condition_cache_state.cpp
```diff
@@ -79,26 +79,128 @@ CacheEntryStats ConditionCacheEntry::ComputeStats(idx_t total_rows) const {
...
+	auto filter_keys = index->GetAll();
+	idx_t removed_count = 0;
+
+	for (const auto &filter_key : filter_keys) {
+		CacheKey key {table_oid, filter_key};
+		string cache_key = MakeCacheKeyString(key);
+		auto entry = cache.Get<ConditionCacheEntry>(cache_key);
+		if (!entry) {
+			// Entry was evicted by LRU, clean up stale index
+			index->Remove(filter_key);
```
dentiny, line 155: since we don't acquire lock when keep removing keys from index, is there any data race concern? For example, query-1 delete index (we're deleting key from index here), query-2 access the same row which should not exist

src/query_condition_cache_state.cpp
dentiny: > If a table's entries are all removed, we could remove it from cached_table_oids as well.

Yes

src/include/query_condition_cache_state.hpp
dentiny: Curious why do we use vector instead of unordered set?

src/query_condition_cache_state.cpp
Andrewtangtang: If the cache entry is evicted by LRU, the entry is gone but the index still holds a stale reference.

src/query_condition_cache_state.cpp
Andrewtangtang: Added `entry->lock` to protect bitvectors access during invalidation.

src/query_condition_cache_state.cpp
dentiny: > the entry is gone but the index still holds a stale reference.

Yes, that's what I mean -- removing key from key index should succeed?

src/include/query_condition_cache_state.hpp
```diff
@@ -78,6 +79,33 @@ struct ConditionCacheEntry : public ObjectCacheEntry {
 	CacheEntryStats ComputeStats(idx_t total_rows) const;
 };
 
+// Per-table index stored in ObjectCache (non-evictable). Tracks which
+// filter_keys have cache entries for a given table, enabling fast
+// invalidation lookup when DML modifies the table.
+struct TableFilterKeyIndex : public ObjectCacheEntry {
+	concurrency::mutex lock;
+	unordered_set<string> filter_keys DUCKDB_GUARDED_BY(lock);
```
dentiny, line 89:
```suggestion
    // When a filter key exists in the set, it means certain key used to be cached for thus table; keys could be evicted.
   // If a key doesn't exist, it's guaranteed not being cached.
	unordered_set<string> filter_keys DUCKDB_GUARDED_BY(lock);
```
let's add some documentations, feel free to modify. :)

src/query_condition_cache_state.cpp
dentiny: > If the cache entry is evicted by LRU, the entry is gone but the index still holds a stale reference.

I think it explains why we could have filter key already exist in the set, but I'm still not sure why remove might non-exist?

Andrewtangtang: I don't think this case can happen in the current code flow. Added a D_ASSERT to catch it if my assumption is wrong.

src/query_condition_cache_state.cpp
```diff
@@ -88,6 +88,7 @@ void TableFilterKeyIndex::Add(const string &filter_key) {
 
 void TableFilterKeyIndex::Remove(const string &filter_key) {
 	concurrency::lock_guard<concurrency::mutex> guard(lock);
+	D_ASSERT(filter_keys.count(filter_key) > 0);
```
dentiny, line 91: Should use always assert and avoid one lookup.. I could followup

dentiny (approved):
Thanks!

PR #55, opened 2026-04-04
Implement cache filter injection with PostOptimizeWalk

peterxcli:
Take the `CacheOptimizerQueryState` that is cooked during pre-optimize from query context, use its data to create the `CacheExpressionFilter` function expression and then push it to `row id column`'s filter

We got up to 150x faster on the  S3-Q2 query of HDFS Logs Benchmark Suite.

image

### DIFF
--- a/src/include/query_condition_cache_filter.hpp
+++ b/src/include/query_condition_cache_filter.hpp
@@ -28,6 +28,8 @@ unique_ptr<FunctionData> ConditionCacheFilterBind(ClientContext &context, Scalar
 unique_ptr<FunctionLocalState> ConditionCacheFilterInit(ExpressionState &state, const BoundFunctionExpression &expr,
                                                         FunctionData *bind_data);
 
+ScalarFunction ConditionCacheFilterFunction();
+
 // Vector-level filter: takes a ROW_ID column as input, looks up the bitvector
 // for that row group + vector index, and returns a constant BOOLEAN for the
 // entire vector. Defaults to true for row groups not in cache.
--- a/src/include/query_condition_cache_filter.hpp
+++ b/src/include/query_condition_cache_filter.hpp
@@ -28,6 +28,7 @@ unique_ptr<FunctionData> ConditionCacheFilterBind(ClientContext &context, Scalar
 unique_ptr<FunctionLocalState> ConditionCacheFilterInit(ExpressionState &state, const BoundFunctionExpression &expr,
                                                         FunctionData *bind_data);
 
+// Factory for the internal scalar filter ConditionCacheFilterFn used by optimizer-injected cache checks.
 ScalarFunction ConditionCacheFilterFunction();
 
 // Vector-level filter: takes a ROW_ID column as input, looks up the bitvector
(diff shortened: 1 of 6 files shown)


dentiny:
I will take a look tonight.

src/include/query_condition_cache_filter.hpp
```diff
@@ -28,6 +28,8 @@ unique_ptr<FunctionData> ConditionCacheFilterBind(ClientContext &context, Scalar
 unique_ptr<FunctionLocalState> ConditionCacheFilterInit(ExpressionState &state, const BoundFunctionExpression &expr,
                                                         FunctionData *bind_data);
 
+ScalarFunction ConditionCacheFilterFunction();
```
dentiny, line 32: nit: Do you mind adding some documentation?

src/query_condition_cache_optimizer.cpp
```diff
@@ -143,9 +146,68 @@ shared_ptr<ConditionCacheEntry> QueryConditionCacheOptimizer::BuildCacheForPredi
...
+	auto entry = state.cache_apply_pending.find(get.table_index);
+	if (entry == state.cache_apply_pending.end()) {
+		return;
+	}
+
+	InjectCacheFilter(context, get, entry->second);
+	state.cache_apply_pending.erase(entry);
+}
+
+void QueryConditionCacheOptimizer::InjectCacheFilter(ClientContext &context, LogicalGet &get,
```
dentiny, line 169: Usually we don't pass `const shared_ptr<>&`, we could pass shared pointer by value here

test/unittest/test_filter.cpp
```diff
@@ -2,11 +2,30 @@
...
+#include "duckdb/planner/expression/bound_function_expression.hpp"
+#include "duckdb/planner/filter/expression_filter.hpp"
 #include "duckdb/planner/expression/bound_reference_expression.hpp"
+#include "duckdb/planner/operator/logical_get.hpp"
 #include "duckdb/storage/statistics/numeric_stats.hpp"
+#include "test_helpers.hpp"
 
 namespace duckdb {
 
+static optional_ptr<LogicalGet> FindLogicalGet(LogicalOperator &op) {
```
dentiny: nit: prefer anonymous namespace

dentiny (approved):
LGTM, don't see any issues, just some nits, feel free to merge.
e2e test wise, do you think we could add sql test to check IO request count / data read from storage?

src/query_condition_cache_optimizer.cpp
peterxcli: my clang tidy says:
```
The parameter 'entry' is copied for each invocation but only used as a const reference; consider making it a const reference (fix available)clang-tidy[performance-unnecessary-value-param](https://clang.llvm.org/extra/clang-tidy/checks/performance/unnecessary-value-param.html)
```

peterxcli: I dont know if it's false positive? not familiar to it...

peterxcli: I'm going to merge first, leaving this for followup

PR #58, opened 2026-04-06
Add PhysicalCacheInvalidator operator

Andrewtangtang:
## Summary

This PR adds `PhysicalCacheInvalidator`, a passthrough physical operator that observes DML data streams (INSERT/DELETE/UPDATE/MERGE) to collect affected row group indices and invalidates them from the condition cache in `OperatorFinalize`. Supports three modes: `ROW_ID`, `INSERT`, and `MERGE`. This operator is not yet wired into query plans; the logical operator and optimizer injection will follow in subsequent PRs.


src/include/physical_cache_invalidator.hpp
```diff
@@ -0,0 +1,56 @@
...
+#include "concurrency/annotated_mutex.hpp"
+#include "concurrency/thread_annotation.hpp"
+
+#include "duckdb/execution/physical_operator.hpp"
+
+namespace duckdb {
+
+enum class CacheInvalidatorMode : uint8_t {
+	// DELETE/UPDATE: observe row IDs at row_id_column_index
+	ROW_ID,
```
dentiny: I'm wondering why don't we use `DELETE`/`UPDATE`/etc?

dentiny (commented):
quick glance... any testing?

src/include/physical_cache_invalidator.hpp
Andrewtangtang: The enum represents the strategy for collecting affected row groups, not the DML type itself. `DELETE` and `UPDATE` both use the same strategy (extract row group indices from row IDs), so naming them separately would be redundant. `MERGE` is a hybrid of ROW_ID + INSERT counting, which doesn't map cleanly to a single DML type either.

dentiny: then `ROW_ID` should be `DELETE`?

src/physical_cache_invalidator.cpp
```diff
@@ -0,0 +1,131 @@
...
+	for (idx_t ii = 0; ii < count; ++ii) {
+		auto sel_idx = row_id_data.sel->get_index(ii);
+		if (!row_id_data.validity.RowIsValid(sel_idx)) {
+			if (track_nulls) {
+				++state.inserted_row_count;
+			}
+			continue;
+		}
+		auto row_id = NumericCast<idx_t>(row_ids[sel_idx]);
+		state.affected_row_groups.insert(row_id / DEFAULT_ROW_GROUP_SIZE);
```
dentiny: seems you're protecting against the concurrent access against `affected_row_groups`
we should be able to reduce the critical section by (1) getting all the affect row groups; (2) insert them together in one batch

src/physical_cache_invalidator.cpp
```diff
@@ -0,0 +1,131 @@
...
+	auto &invalidator_state = input.global_state.Cast<CacheInvalidatorGlobalState>();
+
+	concurrency::lock_guard<concurrency::mutex> guard(invalidator_state.lock);
+	// For INSERT/MERGE modes: compute affected row groups from the inserted row range.
+	// New rows start at pre_insert_row_count and span inserted_row_count rows.
+	if (invalidator_state.inserted_row_count > 0) {
+		idx_t first_rg = pre_insert_row_count / DEFAULT_ROW_GROUP_SIZE;
+		idx_t last_rg = (pre_insert_row_count + invalidator_state.inserted_row_count - 1) / DEFAULT_ROW_GROUP_SIZE;
+		for (idx_t rg = first_rg; rg <= last_rg; ++rg) {
+			invalidator_state.affected_row_groups.insert(rg);
```
dentiny, line 95: At least you can reserve in advance: https://en.cppreference.com/w/cpp/container/unordered_map/reserve.html

test/unittest/test_physical_cache_invalidator.cpp
```diff
@@ -0,0 +1,123 @@
+#include "catch/catch.hpp"
+
+#include "query_condition_cache_state.hpp"
+
+#include "duckdb/main/connection.hpp"
+#include "duckdb/main/database.hpp"
+
+namespace duckdb {
+
+TEST_CASE("PhysicalCacheInvalidator - ROW_ID mode invalidation pattern", "[physical_invalidator]") {
```
dentiny: I think we need to update the name :)

src/include/physical_cache_invalidator.hpp
```diff
@@ -0,0 +1,57 @@
...
+#include "concurrency/annotated_mutex.hpp"
+#include "concurrency/thread_annotation.hpp"
+
+#include "duckdb/execution/physical_operator.hpp"
+
+namespace duckdb {
+
+enum class CacheInvalidatorMode : uint8_t {
+	// Observe row IDs at row_id_column_index to find affected row groups
+	DELETE,
```
dentiny: Let me explain a bit: I understand your previous point that `ROW_ID` mode is used to record affected rows and we will remove it, that's why I suggest to call it `DELETE` or `INVALIDATE`

src/physical_cache_invalidator.cpp
```diff
@@ -0,0 +1,136 @@
...
+// Passthrough operator: forwards input unchanged while collecting affected row groups.
+// Called once per chunk by potentially multiple threads.
+OperatorResultType PhysicalCacheInvalidator::Execute(ExecutionContext &context, DataChunk &input, DataChunk &chunk,
+                                                     GlobalOperatorState &gstate, OperatorState &state) const {
+	chunk.Reference(input);
+	auto &invalidator_state = gstate.Cast<CacheInvalidatorGlobalState>();
+
+	switch (mode) {
+	case CacheInvalidatorMode::DELETE:
+	case CacheInvalidatorMode::UPDATE:
```
dentiny: When we do update (which is a deletion + append from my understanding), shall we increment `inserted_row_count`?
I'm wondering if we could add a unit test to check against row id

test/unittest/include/test_helpers_invalidation.hpp
```diff
@@ -0,0 +1,37 @@
...
+#include "query_condition_cache_state.hpp"
+
+#include "duckdb/catalog/catalog.hpp"
+#include "duckdb/catalog/catalog_entry/duck_table_entry.hpp"
+#include "duckdb/common/string_util.hpp"
+#include "duckdb/main/connection.hpp"
+
+namespace duckdb {
+
+inline idx_t GetTableOid(Connection &con, const string &table_name) {
```
dentiny: I usually don't do `inline` function for those >5 lines of code, the reason is, every time the header file is changed, all translation unit tests depending on it have to recompile.

test/unittest/include/test_helpers_invalidation.hpp
```diff
@@ -0,0 +1,37 @@
...
+	return ConditionCacheStore::GetOrCreate(*con.context);
+}
+
+inline shared_ptr<ConditionCacheEntry> LookupEntry(Connection &con, idx_t table_oid, const string &predicate) {
+	auto store = GetStore(con);
+	return store->Lookup(*con.context, {table_oid, predicate});
+}
+
+inline void BuildCache(Connection &con, const string &table_name, const string &predicate) {
+	auto sql = StringUtil::Format("SELECT * FROM condition_cache_build('%s', '%s')", table_name, predicate);
```
dentiny: Another reason is, if we put the implementation in C++, we don't need to have heavy-loaded header inclusion at header file (which is obviously transitive)

src/physical_cache_invalidator.cpp
Andrewtangtang: I see [PhysicalUpdate::Sink](https://github.com/duckdb/duckdb/blob/796be006a5ca0aa55f8f9720f088789a554b6bc3/src/execution/operator/persistent/physical_update.cpp#L107): non-indexed updates seems call `table.Update()` in-place, while  indexed updates do `table.Delete()` + `table.LocalAppend()` . For indexed `UPDATE`, the next PR
  (CacheInvalidationOptimizer) should use MERGE mode instead of `INVALIDATE` to also track the appended rows. Added a TODO for this.

src/include/physical_cache_invalidator.hpp
```diff
@@ -0,0 +1,59 @@
...
+#include "concurrency/annotated_mutex.hpp"
+#include "concurrency/thread_annotation.hpp"
+
+#include "duckdb/execution/physical_operator.hpp"
+
+namespace duckdb {
+
+enum class CacheInvalidatorMode : uint8_t {
+	// DELETE/UPDATE: observe row IDs at row_id_column_index to find affected row groups.
+	// TODO: Indexed column UPDATE does DELETE + LocalAppend internally. New rows flushed
```
dentiny, line 12: I have question for this: does it apply to cases where indexed rows have been persisted to disk? For example, after all rows have been dumped to disk, we create index for the table.

test/unittest/test_physical_cache_invalidator.cpp
```diff
@@ -0,0 +1,174 @@
...
+		REQUIRE(first_in_rg1 / DEFAULT_ROW_GROUP_SIZE == 1);
+	}
+
+	SECTION("large row IDs map correctly") {
+		idx_t row_id = DEFAULT_ROW_GROUP_SIZE * 10 + 500;
+		REQUIRE(row_id / DEFAULT_ROW_GROUP_SIZE == 10);
+	}
+}
+
+TEST_CASE("PhysicalCacheInvalidator - INSERT range computation", "[physical_invalidator]") {
```
dentiny: Sorry I might be misleading, but I'm asking unit test to check our calculated row group is equal to duckdb one.

test/unittest/test_helpers_invalidation.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+shared_ptr<ConditionCacheEntry> LookupEntry(Connection &con, idx_t table_oid, const string &predicate) {
+	auto store = GetStore(con);
+	return store->Lookup(*con.context, {table_oid, predicate});
+}
+
+void BuildCache(Connection &con, const string &table_name, const string &predicate) {
+	auto sql = StringUtil::Format("SELECT * FROM condition_cache_build('%s', '%s')", table_name, predicate);
+	auto result = con.Query(sql);
+	REQUIRE(result->GetError() == "");
```
dentiny: Should we use `HasError`?
https://github.com/duckdb/duckdb/blob/d196df99216691a9557ad6f2bac075e972a9b29f/src/include/duckdb/main/query_result.hpp#L45

src/physical_cache_invalidator.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+// Collect row group indices from valid row IDs in the given vector.
+// For rows with NULL row IDs (unmatched inserts in MERGE), increment inserted_row_count instead.
+void CollectRowGroups(Vector &row_id_vec, idx_t count, CacheInvalidatorGlobalState &state, bool track_nulls) {
+	UnifiedVectorFormat row_id_data;
+	row_id_vec.ToUnifiedFormat(count, row_id_data);
+	auto row_ids = UnifiedVectorFormat::GetData<row_t>(row_id_data);
+
+	// Collect locally first to minimize critical section
+	unordered_set<idx_t> local_row_groups;
```
dentiny, line 33: You can reserve before emplace: https://en.cppreference.com/w/cpp/container/unordered_set/reserve.html

test/unittest/test_physical_cache_invalidator.cpp
Andrewtangtang: DuckDB's row_groups is private so we can't directly compare against its internal row group metadata.

src/include/physical_cache_invalidator.hpp
Andrewtangtang: I think it applies. update_is_del_and_insert is determined at bind time by checking [GetStorageInfo(context).index_info](https://github.com/duckdb/duckdb/blob/d196df99216691a9557ad6f2bac075e972a9b29f/src/catalog/catalog_entry/table_catalog_entry.cpp#L327). It checks whether the updated columns overlap with any existing
index at query time, regardless of when the index was created.

test/unittest/test_physical_cache_invalidator.cpp
dentiny: got it, thanks!

test/unittest/test_physical_cache_invalidator.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+	// Invalidate all row groups
+	unordered_set<idx_t> all_row_groups = {0, 1};
+	store->RemoveRowGroupsForTable(context, /*table_oid=*/1, all_row_groups);
+
+	REQUIRE(store->Lookup(context, {/*table_oid=*/1, "val = 1"}) == nullptr);
+	REQUIRE_FALSE(store->HasEntriesForTable(context, /*table_oid=*/1));
+}
+
+TEST_CASE("PhysicalCacheInvalidator - row ID to row group mapping", "[physical_invalidator]") {
```
dentiny: if you don't mind, can we delete the test below? think they're just simple maths

dentiny (approved):
Thanks!

PR #60, opened 2026-04-08
Improve predicate key and filter index efficiency

Andrewtangtang:
## Summary

This PR improves efficiency in two areas:
- Pre-compute `ToString()` before sorting conjunction children in `NormalizeExpressionForCacheKey`, avoiding redundant calls during comparison.
- Replace `TableFilterKeyIndex::GetAll()` with `ForEach(callback)` to avoid copying the entire set on every invalidation.

## Related Issues

- None

### DIFF
--- a/src/include/query_condition_cache_state.hpp
+++ b/src/include/query_condition_cache_state.hpp
@@ -129,7 +129,14 @@ struct TableFilterKeyIndex : public ObjectCacheEntry {
 	// Remove a filter key. Assumes the key must appear in the set.
 	void Remove(const string &filter_key);
 	bool IsEmpty();
-	unordered_set<string> GetAll();
+	// Invoke callback for each filter key while holding the lock.
+	template <typename Fn>
+	void ForEach(Fn &&fn) {
+		concurrency::lock_guard<concurrency::mutex> guard(lock);
+		for (const auto &key : filter_keys) {
+			fn(key);
+		}
+	}
 };
 
 // Stored in DuckDB's per-database ObjectCache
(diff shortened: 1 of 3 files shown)


peterxcli (approved):
Looks good.

src/include/query_condition_cache_state.hpp
```diff
@@ -129,7 +129,14 @@ struct TableFilterKeyIndex : public ObjectCacheEntry {
 	// Remove a filter key. Assumes the key must appear in the set.
 	void Remove(const string &filter_key);
 	bool IsEmpty();
-	unordered_set<string> GetAll();
+	// Invoke callback for each filter key while holding the lock.
```
dentiny: Usually it's a bad idea -- because the functor passed in can do anything and hold lock for any long time.

If you really want to save the copy, there're two ways
- Provide a `Take` which transfer the ownership for the whole vector out if applicable
- Or you can have a `vector<shared_ptr<string>>`, which is relatively cheap to copy

dentiny: If you think passing functor works here, please leave a warning that passed in function is supposed to be quick

src/predicate_key_utils.cpp
```diff
@@ -54,14 +54,23 @@ void NormalizeExpressionForCacheKey(Expression &expr) {
...
-	// TODO: ToString() is called O(NlogN) times during sort
+	// Sort conjunction children for canonical ordering.
+	// Pre-compute ToString() for each child.
 	if (expr.GetExpressionClass() == ExpressionClass::BOUND_CONJUNCTION) {
 		auto &conj = expr.Cast<BoundConjunctionExpression>();
-		sort(conj.children.begin(), conj.children.end(),
-		     [](const unique_ptr<Expression> &a, const unique_ptr<Expression> &b) {
-			     return a->ToString() < b->ToString();
-		     });
+		vector<pair<string, idx_t>> keyed;
```
dentiny: please document what's the string, what's the index

dentiny (approved):
Thanks!

PR #22, opened 2026-03-12
Verify and fix query condition cache to pass all duckdb slt

peterxcli:
Closes: https://github.com/dentiny/duckdb-query-condition-cache/issues/21

- skip slow test by default

current fixes:
1. GetExtensionName() override — LogicalCacheInvalidator now returns "query_condition_cache" for serialization
2. Serialization/Deserialization — Implemented Serialize() override and CacheInvalidatorOperatorExtension for plan verification support
3. OperatorExtension registration — Registered CacheInvalidatorOperatorExtension in LoadInternal() with a no-op Bind function
4. __condition_cache_filter registration — Registered as a catalog function so it survives plan deserialization
5. ConditionCacheFilterBind — Returns empty cache entry instead of throwing (needed for deserialization)
6. RemapColumnIndices SEGFAULT fix — Replaced D_ASSERT (no-op in release) with a proper exception for unmapped columns (generated columns)
7. Try-catch in PreOptimizeFunction — Catches exceptions from failed cache builds (e.g., sqrt of negative)
8. Skip DML subplans — PreOptimizeWalk now skips cache building inside DELETE/UPDATE/INSERT/MERGE subtrees
9. Skip small tables — Tables with < 1 row group are not cached
10. Skip indexed tables — Tables with indexes are not cached (index scans are more efficient)
11. HasEntriesForTable guard — Cache invalidation optimizer only injects LogicalCacheInvalidator when the table has cache entries

PR #61, opened 2026-04-09
Add cache invalidation optimizer and logical invalidator

peterxcli:
Passthrough operator injected into DML plans to collect
affected row group indices and invalidate them from the
condition cache in OperatorFinalize. Supports ROW_ID,
INSERT, and MERGE modes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ test/python/__pycache__/
 .claude
 AGENTS.md
 CLAUDE.md
+.cache/
--- a/src/cache_invalidation_optimizer.cpp
+++ b/src/cache_invalidation_optimizer.cpp
@@ -1,0 +1,92 @@
+#include "cache_invalidation_optimizer.hpp"
+
+#include "logical_cache_invalidator.hpp"
+#include "query_condition_cache_functions.hpp"
+
+#include "duckdb/catalog/catalog.hpp"
+#include "duckdb/catalog/catalog_entry/duck_table_entry.hpp"
+#include "duckdb/common/enums/logical_operator_type.hpp"
+#include "duckdb/planner/binder.hpp"
+#include "duckdb/planner/expression/bound_reference_expression.hpp"
+#include "duckdb/planner/operator/logical_delete.hpp"
+#include "duckdb/planner/operator/logical_insert.hpp"
+#include "duckdb/planner/operator/logical_merge_into.hpp"
+#include "duckdb/planner/operator/logical_update.hpp"
+
+namespace duckdb {
+
+CacheInvalidationOptimizer::CacheInvalidationOptimizer() {
+	optimize_function = OptimizeFunction;
+}
+
+void CacheInvalidationOptimizer::WalkPlanForDML(ClientContext &context, unique_ptr<LogicalOperator> &op) {
+	// Recurse into children first
+	for (auto &child : op->children) {
+		WalkPlanForDML(context, child);
+	}
+
+	switch (op->type) {
+	case LogicalOperatorType::LOGICAL_DELETE: {
+		auto &del = op->Cast<LogicalDelete>();
+		auto table_oid = del.table.oid;
+
+		// Copy the row_id expression; it will be resolved during column binding resolution
+		auto row_id_expr = del.expressions[0]->Copy();
+
+		auto invalidator = make_uniq<LogicalCacheInvalidator>(table_oid, std::move(row_id_expr));
+		invalidator->children = std::move(del.children);
+		del.children.clear();
+		del.children.push_back(std::move(invalidator));
+		break;
+	}
+	case LogicalOperatorType::LOGICAL_UPDATE: {
+		auto &upd = op->Cast<LogicalUpdate>();
+		auto table_oid = upd.table.oid;
+
+		// row_id is always the last column in an UPDATE's child output
+		auto row_id_col_idx = upd.children[0]->GetColumnBindings().size() - 1;
+		auto row_id_expr = make_uniq<BoundReferenceExpression>(LogicalType::BIGINT, row_id_col_idx);
+
+		auto invalidator = make_uniq<LogicalCacheInvalidator>(table_oid, std::move(row_id_expr));
+		invalidator->children = std::move(upd.children);
+		upd.children.clear();
+		upd.children.push_back(std::move(invalidator));
+		break;
+	}
+	case LogicalOperatorType::LOGICAL_INSERT: {
+		auto &ins = op->Cast<LogicalInsert>();
+		auto table_oid = ins.table.oid;
+
+		auto &duck_table = ins.table.Cast<DuckTableEntry>();
+		auto pre_insert_rows = duck_table.GetStorage().GetTotalRows();
+
+		auto invalidator = make_uniq<LogicalCacheInvalidator>(table_oid, pre_insert_rows);
+		invalidator->children = std::move(ins.children);
+		ins.children.clear();
+		ins.children.push_back(std::move(invalidator));
+		break;
+	}
+	case LogicalOperatorType::LOGICAL_MERGE_INTO: {
+		auto &merge = op->Cast<LogicalMergeInto>();
+		auto table_oid = merge.table.oid;
+		auto &duck_table = merge.table.Cast<DuckTableEntry>();
+		auto pre_insert_rows = duck_table.GetStorage().GetTotalRows();
+
+		auto row_id_col = merge.row_id_start;
+
+		auto invalidator = make_uniq<LogicalCacheInvalidator>(table_oid, row_id_col, pre_insert_rows);
+		invalidator->children = std::move(merge.children);
+		merge.children.clear();
+		merge.children.push_back(std::move(invalidator));
+		break;
+	}
+	default:
+		break;
+	}
+}
+
+void CacheInvalidationOptimizer::OptimizeFunction(OptimizerExtensionInput &input, unique_ptr<LogicalOperator> &plan) {
+	WalkPlanForDML(input.context, plan);
+}
+
+} // namespace duckdb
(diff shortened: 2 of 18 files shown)


src/cache_invalidation_optimizer.cpp
```diff
@@ -0,0 +1,92 @@
...
+		del.children.push_back(std::move(invalidator));
+		break;
+	}
+	case LogicalOperatorType::LOGICAL_UPDATE: {
+		auto &upd = op->Cast<LogicalUpdate>();
+		auto table_oid = upd.table.oid;
+
+		// row_id is always the last column in an UPDATE's child output
+		auto row_id_col_idx = upd.children[0]->GetColumnBindings().size() - 1;
+		auto row_id_expr = make_uniq<BoundReferenceExpression>(LogicalType::BIGINT, row_id_col_idx);
```
dentiny:
```suggestion
		auto row_id_expr = make_uniq<BoundReferenceExpression>(LogicalType {LogicalTypeId::BIGINT}, row_id_col_idx);
```

.gitignore
```diff
@@ -10,3 +10,4 @@ test/python/__pycache__/
 .claude
 AGENTS.md
 CLAUDE.md
+.cache/
```
dentiny, line 13: hmmm usually hidden directories and files are ignored by git, unless you `git add` explicitly

test/unittest/test_cache_invalidation_optimizer.cpp
```diff
@@ -0,0 +1,133 @@
...
+}
+
+LogicalCacheInvalidator &GetOnlyInvalidatorChild(LogicalOperator &op) {
+	REQUIRE(op.children.size() == 1);
+	REQUIRE(op.children[0]->type == LogicalOperatorType::LOGICAL_EXTENSION_OPERATOR);
+	return op.children[0]->Cast<LogicalCacheInvalidator>();
+}
+
+void VerifyCopiedInvalidator(ClientContext &context, LogicalOperator &plan, CacheInvalidatorMode expected_mode,
+                             idx_t expected_table_oid, idx_t expected_pre_insert_rows,
```
dentiny: nit: let's make it a struct to pass in

src/logical_cache_invalidator.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+// INSERT mode: count rows, no row_id tracking
+LogicalCacheInvalidator::LogicalCacheInvalidator(idx_t table_oid, idx_t pre_insert_row_count)
+    : table_oid(table_oid), mode(CacheInvalidatorMode::INSERT), row_id_column_index(0),
+      pre_insert_row_count(pre_insert_row_count) {
+}
+
+// MERGE mode: track row IDs + count unmatched (inserted) rows
+LogicalCacheInvalidator::LogicalCacheInvalidator(idx_t table_oid, idx_t row_id_column_index, idx_t pre_insert_row_count)
+    : table_oid(table_oid), mode(CacheInvalidatorMode::MERGE), row_id_column_index(row_id_column_index),
```
dentiny:
```suggestion
    : table_oid(table_oid_p), mode(CacheInvalidatorMode::MERGE), row_id_column_index(row_id_column_index_p),
```

dentiny: same elsewhere

src/logical_cache_invalidator.cpp
```diff
@@ -0,0 +1,111 @@
...
+	types = children[0]->types;
+}
+
+string LogicalCacheInvalidator::GetExtensionName() const {
+	return "query_condition_cache";
+}
+
+void LogicalCacheInvalidator::Serialize(Serializer &serializer) const {
+	LogicalExtensionOperator::Serialize(serializer);
+	serializer.WriteProperty(300, "table_oid", table_oid);
```
dentiny, line 58: I'm not too familiar on this, do you mind giving me a lesson how you get the ID? Thanks!

src/logical_cache_invalidator.cpp
```diff
@@ -0,0 +1,111 @@
...
+                                           SQLStatement &statement) {
+	// We don't bind any statements — return empty plan to signal "not handled"
+	return BoundStatement();
+}
+
+CacheInvalidatorOperatorExtension::CacheInvalidatorOperatorExtension() {
+	Bind = CacheInvalidatorBind;
+}
+
+std::string CacheInvalidatorOperatorExtension::GetName() {
```
dentiny:
```suggestion
string CacheInvalidatorOperatorExtension::GetName() {
```

src/logical_cache_invalidator.cpp
```diff
@@ -0,0 +1,111 @@
...
+	serializer.WriteProperty(300, "table_oid", table_oid);
+	serializer.WriteProperty(301, "mode", static_cast<uint8_t>(mode));
+	serializer.WriteProperty(302, "row_id_column_index", row_id_column_index);
+	serializer.WriteProperty(303, "pre_insert_row_count", pre_insert_row_count);
+	serializer.WritePropertyWithDefault(304, "expressions", expressions);
+}
+
+// --- CacheInvalidatorOperatorExtension ---
+
+static BoundStatement CacheInvalidatorBind(ClientContext &context, Binder &binder, OperatorExtensionInfo *info,
```
dentiny:
```suggestion
namespace {
BoundStatement CacheInvalidatorBind(ClientContext &context, Binder &binder, OperatorExtensionInfo *info,
```

test/sql/condition_cache_invalidation.test
```diff
@@ -0,0 +1,35 @@
...
+query I
+SELECT cached_row_groups FROM condition_cache_info('t', 'val = 42');
+----
+2
+
+statement ok
+DELETE FROM t WHERE id < 1000;
+
+query I
+SELECT cached_row_groups FROM condition_cache_info('t', 'val = 42');
```
dentiny: the predicate in the sql test is too easily, can we add some complex ones?

test/unittest/test_cache_invalidation_optimizer.cpp
```diff
@@ -0,0 +1,133 @@
...
+		auto &merge = plan->Cast<LogicalMergeInto>();
+		auto &invalidator = GetOnlyInvalidatorChild(*plan);
+		REQUIRE(invalidator.table_oid == table_oid);
+		REQUIRE(invalidator.mode == CacheInvalidatorMode::MERGE);
+		REQUIRE(invalidator.row_id_column_index == merge.row_id_start);
+		REQUIRE(invalidator.pre_insert_row_count == 3);
+		REQUIRE(invalidator.expressions.empty());
+
+		VerifyCopiedInvalidator(*con.context, *plan, CacheInvalidatorMode::MERGE, table_oid, 3, merge.row_id_start, 0);
+		con.context->transaction.Commit();
```
dentiny, line 136: I'm not too sure what C++ unit tests brings us, over sql tests?

dentiny (commented):
No big comments

test/unittest/test_cache_invalidation_optimizer.cpp
peterxcli: it verify the the optimized plan has expected logical operator with correct argument(property).

src/logical_cache_invalidator.cpp
peterxcli: actually the only rules is not to conflict with used id, [`LogicalOperator`](https://github.com/duckdb/duckdb/blob/7dbb2e646fea939a89f10a55aa98c474cbb0c098/src/storage/serialization/serialize_logical_operator.cpp#L14-L17) owns 100+, [`LogicalExtensionOperator`](https://github.com/duckdb/duckdb/blob/7dbb2e646fea939a89f10a55aa98c474cbb0c098/src/planner/operator/logical_extension_operator.cpp#L23-L26) already owns 200+

spatial ues 300, https://github.com/duckdb/duckdb-spatial/blob/main/src/spatial/operators/spatial_operator_extension.cpp#L28

test/unittest/test_cache_invalidation_optimizer.cpp
peterxcli: while the sql test test the actual result.

src/logical_cache_invalidator.cpp
dentiny: just to be safer, should we use larger numbers than? I'm not familiar with it TBH, I'm worried and not sure whether ID duplication would lead to any issue (pb doesn't allow that)

dentiny (approved):

src/logical_cache_invalidator.cpp
peterxcli: The field-ID namespace is effectively per serialized object, not global across DuckDB. LogicalExtensionOperator first writes extension_name at 200, and deserialization uses that to choose the right extension handler in [logical_extension_operator.cpp](https://github.com/duckdb/duckdb/blob/7dbb2e646fea939a89f10a55aa98c474cbb0c098/src/planner/operator/logical_extension_operator.cpp#L23-L26). So if both duckdb-spatial and query_condition_cache are installed, each serialized operator is still routed to its own extension before reading the extension-specific fields. spatial using 300 and us using 300 is fine.

dentiny: > The field-ID namespace is effectively per serialized object, not global across DuckDB.

Then why does duckdb need it?

peterxcli: `LogicalOperator::Copy` use serde to implement, and some optimizer need to copy the operator, eg. [window self join](https://github.com/duckdb/duckdb/blob/7dbb2e646fea939a89f10a55aa98c474cbb0c098/src/optimizer/window_self_join.cpp#L198), otherwise it would fallback

peterxcli: I think i found this when running the https://github.com/dentiny/duckdb-query-condition-cache/pull/22

PR #63, opened 2026-04-10
Fix test failures from extension-loaded test runs

Andrewtangtang:
## Summary

This PR fixes test failures surfaced by running `make test_release_all` with
the cache config loaded. Three independent issues in the cache optimizer
caused failures:

1. Missing `try-catch` in `PreOptimizeFunction` let binder and evaluation
   errors propagate to queries.
2. Missing `HasIndexes()` guard caused ART index scans to be downgraded to
   sequential scans on cached tables (`InjectCacheFilter` adds a ROW_ID
   filter, but DuckDB only takes the index scan path when
   `filter_set.filters.size() == 1`).
3. `ALIAS()` and other context-dependent functions produced wrong results
   when re-bound via `CheckBinder` round-trip.

Fixes 1 and 3 are workarounds for a deeper issue: `BuildCacheForPredicate`
round-trips through `ToString() -> Parser -> CheckBinder`, but `CheckBinder`
has different semantics than the normal query binder. The correct fix is to
clone `filter.expressions` directly and remap column indices, avoiding
re-binding entirely. Left as a follow-up.

Also cleans up the test config:
- Drops the redundant `on_new_connection` SET (setting already defaults to
  `true`; the SET fails on invalidated/aborted databases used by concurrent
  WAL and ART constraint tests).
- Removes `parquet` from the static extensions list (we don't depend on it,
  and preloading breaks `zstd_crash.test` which requires parquet to be
  absent).

### DIFF
--- a/src/include/predicate_key_utils.hpp
+++ b/src/include/predicate_key_utils.hpp
@@ -18,8 +18,7 @@ string ComputeCanonicalPredicateKey(ClientContext &context, DuckTableEntry &tabl
 // the optimizer path (filter.expressions) to avoid re-binding via CheckBinder.
 string ComputeCanonicalPredicateKey(const vector<unique_ptr<Expression>> &expressions);
 
-// Combine expressions under AND, taking ownership. Single input is returned
-// unchanged.
+// Combine expressions under AND. Single input is returned unchanged.
 unique_ptr<Expression> CombineWithAnd(vector<unique_ptr<Expression>> children);
 
 // Normalize a bound expression tree in-place so equivalent predicates produce
--- a/src/include/predicate_key_utils.hpp
+++ b/src/include/predicate_key_utils.hpp
@@ -6,21 +6,25 @@ namespace duckdb {
 
 class DuckTableEntry;
 
-// Parse a predicate SQL string, bind it against a table using CheckBinder,
-// and return the bound expression. Returns nullptr if predicate_sql is empty
-// or parsing fails.
+// Parse + CheckBinder-bind a predicate SQL against a table. Returns nullptr on
+// empty input or parse failure.
 unique_ptr<Expression> BindPredicate(ClientContext &context, DuckTableEntry &table_entry, const string &predicate_sql);
 
-// Compute a canonical cache key string from a predicate SQL.
-// Parses, binds, normalizes (comparison operand order + conjunction sort),
-// and returns the normalized ToString(). Returns empty string if predicate_sql
-// is empty or parsing fails.
+// Compute a canonical cache key from a predicate SQL string. Use for the table
+// function path where input starts as SQL text.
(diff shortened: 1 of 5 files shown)


peterxcli (approved):
LGTM, let's raise another PR to fix the check binder problem

Andrewtangtang:
Optimizer path now uses plan bindings directly and skips CheckBinder. Table function path (`condition_cache_build` / `condition_cache_info`) still uses CheckBinder for SQL input — both paths produce compatible keys via post-processing. But I think table functions are not the hot path for most of using, so leaving current workarounds to solve them. I will leave a follow-up issue to track.

Andrewtangtang:
cc @peterxcli

src/include/predicate_key_utils.hpp
```diff
@@ -6,21 +6,25 @@ namespace duckdb {
...
 
-// Normalize a bound expression tree in-place for canonical cache key generation.
-// Ensures equivalent predicates produce the same ToString() output:
-//   - Comparison operands: constants moved to the right ("42 = val" -> "val = 42")
-//   - Conjunction children (AND/OR): sorted by ToString() ("b = 2 OR a = 1" -> "a = 1 OR b = 2")
+// Compute a canonical cache key from already-bound filter expressions. Use for
+// the optimizer path (filter.expressions) to avoid re-binding via CheckBinder.
+string ComputeCanonicalPredicateKey(const vector<unique_ptr<Expression>> &expressions);
+
+// Combine expressions under AND, taking ownership. Single input is returned
```
dentiny:
```suggestion
// Combine expressions under AND. Single input is returned unchanged.
```
passing unique pointer implies ownership transfer

src/query_condition_cache_optimizer.cpp
```diff
@@ -80,8 +85,16 @@ void QueryConditionCacheOptimizer::PreOptimizeWalk(ClientContext &context, uniqu
 		return;
 	}
 
-	// TODO: Use std::optional when DuckDB upgrades to C++17 in v2.0
```
dentiny, line 83: TODO still valid?

dentiny (commented):
Feel free to merge, I'm too familiar with this part, will try to catch up this weekend.

src/query_condition_cache_optimizer.cpp
Andrewtangtang: The TODO was on the old ComputePredicateKey helper which had its own if `(!table_ptr)` return `CacheKey{oid, ""}` — using empty string as sentinel. That helper is removed now. The null-table case is already guarded by `PreOptimizeWalk` line 81:(if (!table) return), and `key.filter_key.empty()` at line 98 is kept as a defensive check. So I don't think `std::optional` is still needed anymore.

Andrewtangtang:
Thanks!

PR #64, opened 2026-04-11
Update README.md to reflect latest codebase

peterxcli:
update the `README.md` to provide a comprehensive overview of the `query_condition_cache` DuckDB extension, focusing on its purpose, performance benefits, usage instructions, and development workflow. The new documentation replaces the previous template with clear explanations, benchmark results, and practical examples for both users and developers.

Key documentation improvements:

**Extension overview and performance:**
* Added a detailed description of what `query_condition_cache` does, its use cases (e.g., metrics dashboards, log investigation), and how it improves repeated-query performance by caching predicate results. Included benchmark results showing up to 314.3x speedup and a performance chart.

**Usage instructions:**
* Provided step-by-step SQL examples for loading the extension, creating tables, manually building and inspecting caches, and enabling/disabling automatic cache building via the optimizer.

**Technical explanation:**
* Explained the internal workings of the extension, including predicate canonicalization, vector scanning, and cache storage mechanisms, to help users and developers understand its operation.

**Development and testing:**
* Added clear instructions for building, testing, and running the extension, including sample commands and test invocation.

**Roadmap and future work:**
* Introduced a roadmap section highlighting planned features, such as Parquet predicate scan cache support.


Andrewtangtang:
I would suggest adding a citation or reference.

README.md
```diff
@@ -1,102 +1,78 @@
...
+# DuckDB Query Condition Cache
 
-This repository is based on https://github.com/duckdb/extension-template, check it out if you want to build and ship your own DuckDB extension.
+`query_condition_cache` is a DuckDB extension for repeated-query workloads like **metrics monitoring dashboards**, **log investigation**, and other **recurring read paths** where the same filter shows up again and again.
 
----
+Instead of re-evaluating a predicate across the whole table every time, it remembers which row-group vectors matched before and lets DuckDB skip vectors that are already known to be empty for that predicate.
 
-This extension, QueryConditionCache, allow you to ... <extension_goal>.
+It is especially useful for repeated `LIKE` searches and other expression-heavy filters that are expensive to revisit.
```
dentiny: hmm I'm not sure if we need this sentence

README.md
```diff
@@ -1,102 +1,78 @@
...
-## Building
-### Managing dependencies
-DuckDB extensions uses VCPKG for dependency management. Enabling VCPKG is very simple: follow the [installation instructions](https://vcpkg.io/en/getting-started) or just run the following:
-```shell
-git clone https://github.com/Microsoft/vcpkg.git
-./vcpkg/bootstrap-vcpkg.sh
-export VCPKG_TOOLCHAIN_PATH=`pwd`/vcpkg/scripts/buildsystems/vcpkg.cmake
-```
-Note: VCPKG is only required for extensions that want to rely on it for dependency management. If you want to develop an extension without dependencies, or want to do your own dependency management, just skip this step. Note that the example extension uses VCPKG to build with a dependency for instructive purposes, so when skipping this step the build may not work without removing the dependency.
+On the HDFS_v2 log analytics benchmark (~71M log lines), `query_condition_cache` reached up to **314.3x faster** repeated-query performance. The biggest gains showed up on selective HDFS log investigation queries, where the same predicate can be reused across runs.
```
dentiny: Could you please attach a link to HDFS_v2? TBH I have no idea what it is

README.md
```diff
@@ -1,102 +1,78 @@
...
-```sh
-make test
-```
+## Usage
 
-### Installing the deployed binaries
-To install your extension binaries from S3, you will need to do two things. Firstly, DuckDB should be launched with the
-`allow_unsigned_extensions` option set to true. How to set this will depend on the client you're using. Some examples:
+```sql
+LOAD 'query_condition_cache';
```
dentiny, line 19: Need to download first
```sql
FORCE INSTALL xxx;
```

README.md
```diff
@@ -1,102 +1,78 @@
...
 
-CLI:
-```shell
-duckdb -unsigned
+CREATE TABLE t AS
+SELECT i AS id, i % 100 AS val, 'msg_' || (i % 10) AS msg
```
dentiny, line 23:
```suggestion
-- Predicate cache is automatically populated.
SELECT i AS id, i % 100 AS val, 'msg_' || (i % 10) AS msg
```

README.md
```diff
@@ -1,102 +1,78 @@
...
-```
+### Manual Cache Build
+
+If you already know the filter you care about, build a cache entry directly for that table/predicate pair:
 
-Secondly, you will need to set the repository endpoint in DuckDB to the HTTP url of your bucket + version of the extension
-you want to install. To do this run the following SQL query in DuckDB:
 ```sql
-SET custom_extension_repository='bucket.s3.eu-west-1.amazonaws.com/<your_extension_name>/latest';
+SELECT * FROM condition_cache_build('t', 'val = 42 AND msg LIKE ''%2''');
```
dentiny, line 35: add a comment one for cache construction, one for cache info query

README.md
```diff
@@ -1,102 +1,78 @@
...
+SELECT * FROM condition_cache_info('t', 'val = 42 AND msg LIKE ''%2''');
 ```
-Note that the `/latest` path will allow you to install the latest extension version available for your current version of
-DuckDB. To specify a specific version, you can pass the version instead.
 
-After running these steps, you can install and load your extension using the regular INSTALL/LOAD commands in DuckDB:
+`condition_cache_info` returns:
```
dentiny: No need, just paste the table function result here (column names should be straightforward)

peterxcli:
addressed all review comments

dentiny (approved):
Thanks!

PR #78, opened 2026-04-19
Replace sync cache build with recorder operator

### LINKED ISSUE #75: Incremental Cache Build
the cache expression filter should build the cache too if cache miss when processing the incoming data chunk.

Andrewtangtang:
Closes: #75

## Summary

This PR replaces the synchronous full-table build that the optimizer used to
run on cache miss with a `PhysicalCacheRecorder` operator that observes the
user's own query and builds the cache as a side-effect. The first query on
a new predicate no longer pays an extra full-table scan.

An `observed_vectors` watermark is added to `RowGroupFilter` so the cache
can distinguish "observed empty" (safe to prune) from "not yet observed"
(must scan). Recorder injection is narrow: it only runs on a full cache
miss. Partial refill and scan-skipped row groups are tracked as TODOs.

## Operator Hierarchy

Cache miss — optimizer injects the recorder above `LogicalGet`:

```
LogicalFilter (user's predicate)
  └─ LogicalCacheRecorder                ← new; evaluates predicate, records
       └─ LogicalGet
            ├─ column_ids: [..., ROW_ID]
            ├─ projection_ids: [..., ROW_ID]   ← surfaced for recorder
            └─ table_filters[ROW_ID] = CacheExpressionFilter
```

Cache hit — no recorder, filter only:

```
LogicalFilter (user's predicate)
  └─ LogicalGet
       ├─ column_ids: [..., ROW_ID]
       ├─ projection_ids: [...]           ← ROW_ID read but not projected
       └─ table_filters[ROW_ID] = CacheExpressionFilter
```

Runtime flow for the cache-miss case:

```
PhysicalTableScan
  │ emits chunks (post zone-map / table filter)
  ▼
PhysicalCacheRecorder::Execute
  ├─ compute (rg_idx, vec_idx) from first row_id
  ├─ expr_executor.SelectExpression(chunk)
  └─ RecordChunkObservation -> task-local bitvector + watermark
  │ chunk.Reference(input)   (pass-through)
  ▼
PhysicalFilter / downstream ops

... at query end ...

PhysicalCacheRecorder::OperatorFinalize
  ├─ merge all task-local entries -> destination
  └─ store->Upsert(key, destination)
```

### DIFF
--- a/src/include/query_condition_cache_state.hpp
+++ b/src/include/query_condition_cache_state.hpp
@@ -18,21 +18,23 @@ inline constexpr idx_t BITVECTOR_ARRAY_SIZE = (VECTORS_PER_ROW_GROUP + 63) / 64;
 static_assert(DEFAULT_ROW_GROUP_SIZE % STANDARD_VECTOR_SIZE == 0,
               "DEFAULT_ROW_GROUP_SIZE must be divisible by STANDARD_VECTOR_SIZE");
 
-// Per row-group bitvector: bit[i] = 1 means vector i has at least one qualifying row,
-// for 0 <= i < VECTORS_PER_ROW_GROUP.
-// Backed by array<uint64_t, N> to support configurable row group / vector sizes.
+// Per row-group bitvector + watermark.
+//   matching_vectors[i] = 1 iff vec i observed to have at least one qualifying row.
+//   observed_vectors   = vecs in [0, observed_vectors) are fully observed; anything
+//                        at or above this index is unknown and must be scanned.
 struct RowGroupFilter {
 	array<uint64_t, BITVECTOR_ARRAY_SIZE> matching_vectors = {};
+	idx_t observed_vectors = 0;
 
 	RowGroupFilter() = default;
 
-	// Construct from vector indices that contain at least one qualifying row.
-	// Each index must be in [0, VECTORS_PER_ROW_GROUP). Duplicates are allowed.
+	// Leaves observed_vectors at 0; set it explicitly if you need pruning semantics.
 	explicit RowGroupFilter(const vector<idx_t> &qualifying_vectors);
 
 	void SetVector(idx_t vector_index);
 	bool VectorHasRows(idx_t vector_index) const;
 	bool IsEmpty() const;
+	// OR of matching_vectors, max of observed_vectors.
 	void MergeFrom(const RowGroupFilter &other);
 };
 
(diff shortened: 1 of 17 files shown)


src/include/query_condition_cache_state.hpp
```diff
@@ -18,21 +18,23 @@ inline constexpr idx_t BITVECTOR_ARRAY_SIZE = (VECTORS_PER_ROW_GROUP + 63) / 64;
...
-// Per row-group bitvector: bit[i] = 1 means vector i has at least one qualifying row,
-// for 0 <= i < VECTORS_PER_ROW_GROUP.
-// Backed by array<uint64_t, N> to support configurable row group / vector sizes.
+// Per row-group bitvector + watermark.
+//   matching_vectors[i] = 1 iff vec i observed to have at least one qualifying row.
+//   observed_vectors   = vecs in [0, observed_vectors) are fully observed; anything
+//                        at or above this index is unknown and must be scanned.
 struct RowGroupFilter {
 	array<uint64_t, BITVECTOR_ARRAY_SIZE> matching_vectors = {};
+	idx_t observed_vectors = 0;
```
peterxcli: I think the watermark management is not quite good for cache invalidation.
how about we store another bitmask to record if the vector is observed?

```cpp
struct RowGroupFilter {
	array<uint64_t, BITVECTOR_ARRAY_SIZE> matching_vectors = {};
    array<uint64_t, BITVECTOR_ARRAY_SIZE> observed = {};
```

Andrewtangtang: agree.

src/query_condition_cache_optimizer.cpp
```diff
@@ -103,62 +122,30 @@ void QueryConditionCacheOptimizer::PreOptimizeWalk(ClientContext &context, uniqu
...
-		// to avoid blocking the first query.
-		entry = BuildCacheForPredicate(context, filter.expressions, get);
-		if (entry) {
-			store->Upsert(context, key, entry);
+		// Upsert the empty entry up front so the cache filter's bind_data and the recorder's
+		// Finalize Lookup resolve to the same shared_ptr.
+		entry = make_shared_ptr<ConditionCacheEntry>();
+		store->Upsert(context, key, entry);
+
+		// TODO: Also inject the recorder on a partial cache hit so the watermark can
```
peterxcli, line 130: I think we already removed the watermark?

src/query_condition_cache_optimizer.cpp
```diff
@@ -212,13 +206,57 @@ void QueryConditionCacheOptimizer::InjectCacheFilter(ClientContext &context, Log
 	                             make_uniq<CacheExpressionFilter>(std::move(filter_expr), entry));
 }
 
+void QueryConditionCacheOptimizer::InjectCacheRecorder(ClientContext &context, unique_ptr<LogicalOperator> &plan,
```
peterxcli, line 209: can you use plain words to explain the logic of this function, espically the `rowid_chunk_idx` and `rowid_column_ids_pos`

src/query_condition_cache_optimizer.cpp
```diff
@@ -212,13 +206,57 @@ void QueryConditionCacheOptimizer::InjectCacheFilter(ClientContext &context, Log
...
+}
+
 void QueryConditionCacheOptimizer::OptimizeFunction(OptimizerExtensionInput &input, unique_ptr<LogicalOperator> &plan) {
 	if (!IsSettingEnabled(input.context)) {
 		return;
 	}
 
 	auto query_state = input.context.registered_state->Get<CacheOptimizerQueryState>(CacheOptimizerQueryState::NAME);
-	if (!query_state || query_state->cache_apply_pending.empty()) {
+	if (!query_state || (query_state->cache_apply_pending.empty() && query_state->cache_recorder_pending.empty())) {
```
peterxcli, line 259: our disscusion is to keep keep the recorder anyway?  and thus I think we can remove the two phase optimizer(query context)? because we aren't going to do any cache lookup thing when plan rewriting, we do all lookup and upsert things inside physical recorder operator and the condition cache filter expression when they are processing the incoming data chunk?

src/physical_cache_recorder.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+	auto &rowid_vec = input.data[rowid_column_index];
+	UnifiedVectorFormat rowid_data;
+	rowid_vec.ToUnifiedFormat(input.size(), rowid_data);
+	auto rowids = UnifiedVectorFormat::GetData<row_t>(rowid_data);
+
+	auto first_idx = rowid_data.sel->get_index(0);
+	if (!rowid_data.validity.RowIsValid(first_idx)) {
+		return OperatorResultType::NEED_MORE_INPUT;
+	}
```
peterxcli, line 71: is this case possible? if this happened means no rows in the vector is valid, and I though all invalid vector wont flow into here?

peterxcli (commented):
We should think this through:

1. When the recorder sees a vector that hasn’t been observed yet, it should upsert both the `observed` and `matching_vectors` bitmasks.
2. Do we actually only need `observed` per row group? The recorder could upsert all cache entries for every vector it sees, then mark that row group as observed.
3. The query cache expression filter should only skip a vector if that vector’s row group is marked as observed in the cache entry.

PR #81, opened 2026-04-22
Add condition cache query profiling

### LINKED ISSUE #80: Per Query Stats
We should extend observability beyond global cache stats.

ClickHouse exposes:
- cache contents via system.query_condition_cache
- per-query hits/misses via system.query_log ProfileEvents
- trace-level pruning details like granules dropped / marks read

Iceberg's ScanReport is also a good model: it reports planning duration and scanned/skipped manifests/files. For our future Parquet/Iceberg work, the metrics should generalize over:
table -> data_unit -> data_range -> predicate -> bitmask/result.

Proposal:
1. Keep condition_cache_stats() for global counters:
   entry_count, table_count, memory, lookup_hit/miss, build_count, applied_count,
   effective_hit_count, invalidation/eviction counters.

2. Add condition_cache_entries():
   one row per cache entry, similar to ClickHouse system.query_condition_cache:
   table, predicate_hash, optional predicate_text, memory, cached row groups,
   qualifying vectors, total vectors, selectivity, created/last-used timestamps.

3. Add condition_cache_last_query() and condition_cache_query_history(n):
   one row per scan/probe in recent queries:
   query_seq, scan_id, table, predicate_hash, status, skip_reason, lookup_hit,
   built_inline, build_time_us, applied, row_groups/vectors pruned,
   approx rows pruned, uncached pass-through vectors.

4. Add EXPLAIN ANALYZE / JSON profiling extra_info for cached scans:
   Query Condition Cache: status, predicate_hash, cached_row_groups,
   qualifying_vectors, runtime_vectors_pruned, approx_rows_pruned.

Important: distinguish lookup hit from effective hit.
A lookup hit that prunes zero vectors is useful for debugging but should not be
counted as an effective cache benefit.

---

- create a query-local profile object, e.g. shared_ptr<ConditionCacheProfileInfo>
- attach it to the scan bind data
- set get.function.dynamic_to_string = ConditionCacheDynamicToString

The practical storage place is a custom bind-data type derived from TableScanBindData, because dynamic_to_string only gets bind_data, local_state, and global_state

```cpp
struct ConditionCacheProfileInfo {
	atomic<bool> initial_hit {false};
	atomic<bool> built_this_query {false};
	atomic<idx_t> vectors_checked {0};
	atomic<idx_t> vectors_pruned {0};
	atomic<idx_t> rows_pruned {0};
};

struct ConditionCacheTableScanBindData : public TableScanBindData {
	ConditionCacheTableScanBindData(TableCatalogEntry &table, shared_ptr<ConditionCacheProfileInfo> profile_p)
	    : TableScanBindData(table), profile(std::move(profile_p)) {
	}

	shared_ptr<ConditionCacheProfileInfo> profile;

	unique_ptr<FunctionData> Copy() const override {
		auto result = make_uniq<ConditionCacheTableScanBindData>(table, profile);
		result->is_index_scan = is_index_scan;
		result->is_create_index = is_create_index;
		result->column_ids = column_ids;
		result->order_options = order_options ? make_uniq<RowGroupOrderOptions>(*order_options) : nullptr;
		return std::move(result);
	}
};

static InsertionOrderPreservingMap<string>
ConditionCacheDynamicToString(TableFunctionDynamicToStringInput &input) {
	auto &bind = input.bind_data->Cast<ConditionCacheTableScanBindData>();
	auto &p = *bind.profile;

	InsertionOrderPreservingMap<string> result;
	result["Condition Cache"] = p.built_this_query ? "MISS -> BUILT" : (p.initial_hit ? "HIT" : "MISS");
	result["Cache Vectors Checked"] = to_string(p.vectors_checked.load());
	result["Cache Vectors Pruned"] = to_string(p.vectors_pruned.load());
	result["Cache Rows Pruned"] = to_string(p.rows_pruned.load());
	return result;
}
```

Then update that shared object from:

- ConditionCacheFilterFn(...) on the hit/apply path: query_condition_cache_filter.cpp (line 42)
- PhysicalCacheRecorder::Execute(...) and OperatorFinalize(...) on the miss/build path if you take PR #78’s approach

peterxcli:
Closes: https://github.com/dentiny/duckdb-query-condition-cache/issues/80

Introduce profiling, allowing cache status and statistics to be surfaced in `EXPLAIN ANALYZE` output and JSON profiling.

* Added a `ConditionCacheProfileInfo` to store profiling data (cache status, predicate hash, cached row groups, qualifying vectors, etc.) for each condition cache entry and propagate it through query planning and execution. 
* Modified the optimizer (`QueryConditionCacheOptimizer`) to create and attach `ConditionCacheProfileInfo` to each cache entry, and propagate this information through the optimizer state and into the logical plan. 
* Introduced `ConditionCacheTableScanBindData` to wrap table scan bind data with profiling info, and updated the logical get node and filter binding to use and copy this profiling data. 
* Implemented `ConditionCacheDynamicToString` to surface cache profiling information in `EXPLAIN ANALYZE` and JSON profiling output, including status, predicate hash, and statistics.

### DIFF
--- a/src/include/query_condition_cache_state.hpp
+++ b/src/include/query_condition_cache_state.hpp
@@ -60,6 +60,17 @@ struct CacheEntryStats {
 	idx_t total_row_groups;
 };
 
+struct ConditionCacheProfileInfo {
+	atomic<bool> initial_lookup_hit {false};
+	atomic<bool> built_this_query {false};
+
+	string predicate_hash;
+	idx_t cached_row_groups = 0;
+	idx_t qualifying_vectors = 0;
+	idx_t total_vectors = 0;
+	idx_t total_rows = 0;
+};
+
 // A single cache entry: the bitvectors for one (table, predicate) combination.
 struct ConditionCacheEntry : public ObjectCacheEntry {
 	static string ObjectType() {
--- a/src/include/query_condition_cache_filter.hpp
+++ b/src/include/query_condition_cache_filter.hpp
@@ -3,18 +3,31 @@
 #include "query_condition_cache_state.hpp"
 
 #include "duckdb/function/scalar_function.hpp"
+#include "duckdb/function/table/table_scan.hpp"
 #include "duckdb/planner/expression.hpp"
 #include "duckdb/planner/filter/expression_filter.hpp"
 #include "duckdb/storage/statistics/numeric_stats.hpp"
 
 namespace duckdb {
 
+struct ConditionCacheTableScanBindData : public TableScanBindData {
+	ConditionCacheTableScanBindData(TableCatalogEntry &table, shared_ptr<ConditionCacheProfileInfo> profile_info_p);
+
+	shared_ptr<ConditionCacheProfileInfo> profile_info;
+
+	unique_ptr<FunctionData> Copy() const override;
+};
+
+InsertionOrderPreservingMap<string> ConditionCacheDynamicToString(TableFunctionDynamicToStringInput &input);
+
 // Holds a shared_ptr to the ConditionCacheEntry so the filter function
 // and CheckStatistics can query the cache via ConditionCacheEntry's thread-safe API.
 struct ConditionCacheFilterBindData : public FunctionData {
 	shared_ptr<ConditionCacheEntry> cache_entry;
+	shared_ptr<ConditionCacheProfileInfo> profile_info;
 
-	explicit ConditionCacheFilterBindData(shared_ptr<ConditionCacheEntry> entry);
+	ConditionCacheFilterBindData(shared_ptr<ConditionCacheEntry> entry,
+	                             shared_ptr<ConditionCacheProfileInfo> profile_info_p = nullptr);
 
 	unique_ptr<FunctionData> Copy() const override;
 	bool Equals(const FunctionData &other_p) const override;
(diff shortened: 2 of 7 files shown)


dentiny (commented):
For this PR, should we have sql tests (with EXPLAIN ANALYZE)?

peterxcli:
> For this PR, should we have sql tests (with EXPLAIN ANALYZE)?

yes, we already have it
```sql
query II
EXPLAIN (ANALYZE, FORMAT JSON) SELECT count(*) FROM t WHERE id < 3000;
----
analyzed_plan	<REGEX>:.*"Condition Cache": "MISS -> BUILT".*"Condition Cache Predicate Hash": "[0-9]+".*"Condition Cache Cached Row Groups": "5".*"Condition Cache Qualifying Vectors": "2/245".*
```

src/include/query_condition_cache_state.hpp
```diff
@@ -60,6 +60,17 @@ struct CacheEntryStats {
 	idx_t total_row_groups;
 };
 
+struct ConditionCacheProfileInfo {
+	atomic<bool> initial_lookup_hit {false};
```
dentiny, line 64: comment needed

src/include/query_condition_cache_state.hpp
```diff
@@ -60,6 +60,17 @@ struct CacheEntryStats {
...
+	atomic<bool> built_this_query {false};
```
dentiny, line 65: comment needed, also we should name the variable as noun_verb_ed

src/include/query_condition_cache_state.hpp
```diff
@@ -60,6 +60,17 @@ struct CacheEntryStats {
 	idx_t total_row_groups;
 };
 
+struct ConditionCacheProfileInfo {
```
dentiny, line 63: comment needed, does the struct indicate a cache entry, singular/composite predicate, or a user query, a user connection

dentiny:
Could you please paste the overall ANALYZE output in the PR description?

dentiny (commented):
quick review

PR #87, opened 2026-07-14
Fix cache filter alignment across row groups

Andrewtangtang:
## Summary

  The cache originally assumed that every DuckDB scan vector aligned with a
  cache boundary derived from `STANDARD_VECTOR_SIZE`. Based on that assumption,
  the filter checked only the first row ID and reused its cache decision for the
  entire scan vector.

  This assumption fails when a physical row group contains fewer than
  `DEFAULT_ROW_GROUP_SIZE` rows. The next row group can begin between two cache
  boundaries, causing one scan vector to overlap two cache vectors. Reusing only
  the first decision can then incorrectly discard qualifying rows.

  This change checks the first and last row IDs of each scan vector. When the
  vector crosses a cache boundary, each row receives the decision for its
  corresponding cache vector.

  Regression tests cover misaligned scan vectors and automatic cache builds.

  ## Related issues

  None.

### DIFF
--- a/test/unittest/test_filter.cpp
+++ b/test/unittest/test_filter.cpp
@@ -100,10 +100,10 @@ TEST_CASE("ConditionCacheFilterFn handles scan vectors crossing cache-vector bou
 	executor.ExecuteExpression(input, result);
 
 	for (idx_t idx = 0; idx < 512; idx++) {
-		REQUIRE(result.GetValue(idx).GetValue<bool>() == false);
+		REQUIRE(!result.GetValue(idx).GetValue<bool>());
 	}
 	for (idx_t idx = 512; idx < STANDARD_VECTOR_SIZE; idx++) {
-		REQUIRE(result.GetValue(idx).GetValue<bool>() == true);
+		REQUIRE(result.GetValue(idx).GetValue<bool>());
 	}
 }
 
--- a/test/unittest/test_filter.cpp
+++ b/test/unittest/test_filter.cpp
@@ -2,11 +2,17 @@
 #include "query_condition_cache_filter.hpp"
 #include "query_condition_cache_state.hpp"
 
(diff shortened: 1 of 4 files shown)


test/unittest/test_filter.cpp
```diff
@@ -67,6 +73,40 @@ TEST_CASE("CacheExpressionFilter - CheckStatistics", "[query_condition_cache]")
...
+	const auto row_ids = FlatVector::GetData<int64_t>(input.data[0]);
+	for (idx_t idx = 0; idx < STANDARD_VECTOR_SIZE; idx++) {
+		row_ids[idx] = NumericCast<int64_t>(1536 + idx);
+	}
+
+	Vector result(LogicalType {LogicalTypeId::BOOLEAN});
+	executor.ExecuteExpression(input, result);
+
+	for (idx_t idx = 0; idx < 512; idx++) {
+		REQUIRE(result.GetValue(idx).GetValue<bool>() == false);
```
dentiny:
```suggestion
		REQUIRE(!result.GetValue(idx).GetValue<bool>());
```

test/unittest/test_filter.cpp
```diff
@@ -67,6 +73,40 @@ TEST_CASE("CacheExpressionFilter - CheckStatistics", "[query_condition_cache]")
...
+	}
+	for (idx_t idx = 512; idx < STANDARD_VECTOR_SIZE; idx++) {
+		REQUIRE(result.GetValue(idx).GetValue<bool>() == true);
```
dentiny:
```suggestion
		REQUIRE(result.GetValue(idx).GetValue<bool>());
```

dentiny (approved):
Thanks! just two nits

dentiny (approved):

